PR #18, opened 2022-11-30
feat: add draco compression support

rYuuk:
## Description

#### Added

-   List your additions and new features here.

#### Updated

-   List your updates and changes here.

#### Removed

-   List what is removed here.

## How to Test

## Checklist

-   [ ] Changelog is updated.

## QA Testing

## Details


Editor/AvatarConfigEditor.cs
```diff
@@ -8,12 +9,29 @@ namespace ReadyPlayerMe.AvatarLoader.Core
     public class AvatarConfigEditor : Editor
     {
         private AvatarConfig avatarConfigTarget;
+        private SerializedProperty userDracoCompressionField;
```
richieRPM, line 12: Do we need this SerializedProperty... or can we access the the property directly via the avatarConfigTarget assigned on line 16?

rYuuk: Yep, with directly we don't get the value in field.

rYuuk: I was wrong it does, but I had to check the value before OnInspectorGUI is called, which I do in OnEnable.

Editor/AvatarConfigEditor.cs
```diff
@@ -8,12 +9,29 @@ namespace ReadyPlayerMe.AvatarLoader.Core
...
+                if (EditorUtility.DisplayDialog("Read Player Me", "Do you want to install Draco Compression Unity Package: com.atteneder.draco ?", "Ok", "Cancel"))
+                {
+                    ModuleInstaller.AddModule(ModuleList.DracoCompressionModule);
+                }
+            }
+        }
+
+        private void OnEnable()
+        {
+            userDracoCompressionField = serializedObject.FindProperty("UseDracoCompression");
```
HarrisonHough, line 34: MInor but some of these strings could be moved to constants

Editor/ModuleInstaller.cs
```diff
@@ -70,17 +79,17 @@ private static PackageCollection GetPackageList()
...
                 Debug.Log("Error: " + listRequest.Error.message);
                 return null;
             }
-            
+
             return listRequest.Result;
         }
 
-        private static void AddModule(string name)
+        private static void AddPackage(string name)
```
HarrisonHough, line 92: Perhaps
```suggestion
        private static void AddRequestPackage(string name)
```

HarrisonHough (approved):
Some minor things in the comments, but otherwise all good :)

PR #19, opened 2022-12-01
Settings and Module Updates

srcnalt:
### Changes
- Settings file structure changed to help other packages use it
- Module update logic cleaned up
- Package json updated

### How to test
- Import the core package from https://github.com/readyplayerme/Unity-Core.git#feature/webview-updates
- Make a release for WebView and increase the version number
- Click Ready Player Me > Check for Updates button
- Expect package to be updated


HarrisonHough (changes requested):
it seems that whenever you play the scene in editor, or import samples, or compile scripts, the updater runs again and says "installing Avatar Loader". I'm guessing that's not expected behaviour?

Editor/Module Management/ModuleInstaller.cs
```diff
@@ -1,64 +1,47 @@
...
 {
-    [InitializeOnLoad]
-    public class ModuleInstaller 
+    public static class ModuleInstaller 
     {
         private const string PROGRESS_BAR_TITLE = "Ready Player Me";
 
         public static Action ModuleInstallComplete;
 
 #if !DISABLE_AUTO_INSTALLER
```
HarrisonHough, line 16: I think this #if isn't really necessary as we have it in the EntryPoint.cs.

Editor/Module Management/ModuleInstaller.cs
```diff
@@ -1,64 +1,47 @@
...
-        static ModuleInstaller()
+        public static void Init()
         {
             var listRequest = Client.List(true);
             while (!listRequest.IsCompleted)
-                Thread.Sleep(100);
+                Thread.Sleep(20);
             if (HasAnyMissingModule())
             {
-                EditorApplication.update += InstallModules;
```
HarrisonHough, line 27: need to really test this change as I had it in combination with the EditorApplication.Update because there was some issues where this would run before the assets were ready (mostly related to the Scriptable Objects creation). it was causing null references. 
It might be fine now though with all the latest changes but just mentioning here for awareness

Editor/Module Management/ModuleInstaller.cs
```diff
@@ -1,64 +1,47 @@
...
 
             var count = ModuleList.Modules.Length;
-            for (var i = 0; i < count; i++)
+            for(var i = 0; i < count; i++)
             {
+                var packages = GetPackageList();
+                
                 var module = ModuleList.Modules[i];
+                
                 EditorUtility.DisplayProgressBar(PROGRESS_BAR_TITLE, $"Installing module {module.name}", i * count * 0.1f + 0.1f);
```
HarrisonHough, line 41: minor thing but I have no idea what these magic numbers to 0.1f + 01f

Editor/Module Management/ModuleInstaller.cs
```diff
@@ -69,34 +52,51 @@ private static void InstallModules()
...
                 Debug.Log("Error: " + listRequest.Error.message);
                 return null;
             }
-
+            
             return listRequest.Result;
         }
 
-        private static void AddRequestPackage(string name)
+        public static void AddModule(string name)
```
HarrisonHough, line 82: Personally, I prefer AddRequestPackage as it's more clear what it does. 
In general, I think it's kinda weird to have both an AddModule and InstallModule, As the name implies the same thing basically.

Editor/Module Management/ModuleList.cs
```diff
@@ -1,28 +1,27 @@
...
-                branch = ""
+                branch = "develop"
             },
-            new ModuleInfo
+            new ModuleInfo 
             {
                 name = "com.readyplayerme.webview",
                 gitUrl = "https://github.com/readyplayerme/Unity-WebView.git",
-                branch = ""
+                branch = "feature/webview-updates"
```
HarrisonHough, line 23: make sure to remove this before merging (or after)

Editor/Module Management/ModuleUpdater.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                // TODO: Bring analytics here
+                Debug.LogWarning("Update skipped.");
+            }
+        }
+
+        private static void Update(string name, string url, Version current, Version latest)
+        {
+            RemoveRequest removeRequest = Client.Remove(name);
+            while (!removeRequest.IsCompleted) Thread.Sleep(20);
```
HarrisonHough, line 98: we use this Sleep(20) multiple times, perhaps we should store the 20 in a constant field for reuse?

PR #22, opened 2022-12-05
Small fix to prevent installer from running everytime Unity recompiles

HarrisonHough:
## Description

-   previously the module installer would check and install missing modules everytime you recompile. With this update it will now check only once per session. So if you remove a module, EG glTFast it won't auto reinstall, until you reopen Unity. 

#### Updated

-   Module installer now has an Init method that is only called once per session via EntryPoint.cs
- also moved the `#if !DISABLE_AUTO_INSTALLER`, instead of being wrapped around the whole moduleInstaller class it's now only wrapped around ModuleInstalelr.Init() inside Entry point. This should also help with previous issues of ModuleInstaller being missing inside Unity-SDK project. 

## How to Test

-   OPen a project, import core module https://github.com/readyplayerme/Unity-Core.git#feature/installer-fix
- then in editor remove glTFast, and check that it doesn't auto reimport until you restart Unity. 

## Checklist

-   [ ] Changelog is updated.

## QA Testing

## Details


rYuuk:
I made some changes in ModuleInstaller in draco compression, you might need to update the branch from develop.

srcnalt (approved):

PR #23, opened 2022-12-05
Module Installer Updates

srcnalt:


HarrisonHough (approved):
Looks good! Need to test more on different platforms and Unity versions but find to merge I think

PR #24, opened 2022-12-06
fix null exception when loading settings

rYuuk:
## Description

Fixed null exception when loading instance of settings through settings asset.

## How to Test

Verify if there is no null exception while opening settings window

### DIFF
--- a/Configurations/Avatar Config Low.asset
+++ b/Configurations/Avatar Config Low.asset
@@ -18,4 +18,5 @@ MonoBehaviour:
   TextureSizeLimit: 256
   UseHands: 0
   UseDracoCompression: 0
-  MorphTargets: []
+  MorphTargets:
+  - none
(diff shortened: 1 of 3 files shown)


HarrisonHough (approved):
It works for most purposes, however for Unity-SDK project it would crash/breack if we tried to build with WebView because WebViewPanel will try to load from resources to get the subdomain 
image

srcnalt:
Shall we make things so that building from Unity-SDK works? I always import into a project and build to go tru user experience.

HarrisonHough:
> Shall we make things so that building from Unity-SDK works? I always import into a project and build to go tru user experience.

an easy workaround is to just create a RPM settings object inside Unity-SDK/Assets/Resources/.....

srcnalt:
> > Shall we make things so that building from Unity-SDK works? I always import into a project and build to go tru user experience.
> 
> an easy workaround is to just create a RPM settings object inside Unity-SDK/Assets/Resources/Settings.....

Yes tbh,m getting from same source just makes sense to me, both for settings and configs, I'd rather have them in a resources folder in Unity-SDK as well

srcnalt (approved):

PR #26, opened 2022-12-08
[3639140773] Add scripts from core

rYuuk:
## Description

-  Move avatar loader specific scripts from core to avatar loader.

#### Removed

- All data classes specific to AvatarLoader
- Editor settings classes specific to AvatarLoader

## How to Test

-  Best way to test is to check [this](https://github.com/readyplayerme/Unity-SDK/pull/16) PR in Unity-SDK repo 

## Checklist

-   [ ] Changelog is updated.

## QA Testing

## Details


Editor/EditorAssetLoader.cs
```diff
@@ -22,18 +22,15 @@ public static void CreateSettingsAssets()
             AssetDatabase.Refresh();
             CreateAvatarConfigAssets();
             CreateAvatarLoaderSettings();
-            CreateReadyPlayerMeSettings();
+            CreatePartnerSubDomainSetting();
```
HarrisonHough, line 25: SubDomain should be Subdomain. 

Kind of a side note but we could perhaps use PartnerName. or a PartnerSettings object with property Name. However not sure if there would ever be any other "Partner Settings" in future :D 

anyways PartnerSubdomainSettings is fine

HarrisonHough (changes requested):
Looks good, one small thing in the comments (just need to update SubDomain to Subdomain)

srcnalt:
I get error messages for `using ReadyPlayerMe.AvatarLoader;` in Core. Core loads AvatarLoader, and until it is loaded does not have any knowledge of its code base. So we should not mention any code of other modules.

You can import Core into an empty Unity project and see if they show up. 
image

HarrisonHough:
> I get error messages for `using ReadyPlayerMe.AvatarLoader;` in Core. Core loads AvatarLoader, and until it is loaded does not have any knowledge of its code base. So we should not mention any code of other modules.
> 
> You can import Core into an empty Unity project and see if they show up. image

yep this include can be removed, probably just something you missed 

image

HarrisonHough (approved):
Looks good now

PR #33, opened 2023-01-10
Cleanup

srcnalt:
Folder structure and some classes are refactored.


Editor/UI/EditorWindows/EditorWindowBase.cs
```diff
@@ -54,34 +59,33 @@ private void LoadAssets()
...
-            if (webButtonStyle == null)
+            webButtonStyle ??= new GUIStyle(GUI.skin.button)
             {
-                webButtonStyle = new GUIStyle(GUI.skin.button);
-                webButtonStyle.fontSize = 12;
-                webButtonStyle.fixedWidth = 149;
-                webButtonStyle.fixedHeight = ButtonHeight;
-                webButtonStyle.padding = new RectOffset(5, 5, 5, 5);
-            }
+                fontSize = 12,
```
HarrisonHough, line 76: minor thing but we have font size hardcoded to 12 in two places, we could instead store it in a constant.

Editor/UI/EditorWindows/EditorWindowBase.cs
```diff
@@ -12,35 +12,40 @@ namespace ReadyPlayerMe.Core.Editor
...
-        /// 
-        /// </summary>
         private void LoadAssets()
         {
-            banner ??= new Banner();
+            banner ??= new Banner(editorWindowName);
+            
+            footer ??= new Footer(editorWindowName);
+            
+            // TODO: move this into avatar loader settings window
```
HarrisonHough, line 38: para-phrasing a quote from Pablo "TODO's never get done" :D probably we could remove

HarrisonHough (approved):
some minor things in comments but all good otherwise

Editor/UI/EditorWindows/EditorWindowBase.cs
srcnalt: shit forgot this here, will take a look now

PR #37, opened 2023-01-27
[3739587991] Feature/optional logging

HarrisonHough:
## Description

-   Added setting either enable or disable the RPM logger in RPM Settings window under "Other" section. It uses a "ProjectPref" to keep its state in between sessions.              

#### Added

-   Logging enabled setting for SDK logger

## How to Test

## Checklist

-   [ ] Changelog is updated.

## QA Testing

## Details

### DIFF
--- a/Runtime/Logger/SDKLogger.cs
+++ b/Runtime/Logger/SDKLogger.cs
@@ -1,4 +1,5 @@
-﻿using UnityEngine;
+﻿using UnityEditor;
+using UnityEngine;
 
 namespace ReadyPlayerMe.Core
 {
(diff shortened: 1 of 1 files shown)


rYuuk:
I would suggest adding an option in settings instead which they could enable and disable easily. As most user will not use it if its a compiler directive which they would need to add in build settings and this will stay disabled forever. 
Also you need to fix the tests for logs as well.

HarrisonHough:
> I would suggest adding an option in settings instead, so developers could enable and disable it easily. As most user will not use it if its a compiler directive which they would need to add in build settings and this will stay disabled forever. Also you need to fix the tests for logs as well.

Yep, good point will add to the settings. Perhaps a combination of settings and prefs to make it stay enabled/disabled.

HarrisonHough:
Updated, not an extra setting in settings window 
image

rYuuk (approved):
🟢 Great Job and thanks for making the requested change 💯

PR #46, opened 2023-02-16
fix: force create new settings asset

HarrisonHough:
## Description

-  there is currently a number of issues  cause if the core settings asset is not there (or has been deleted) this update fixed that and forces it to be created if it does not exist. 

#### Updated

-   CoreSettingsHandler now creates core settings if missing

## How to Test

## Checklist

-   [ ] Changelog is updated.

## QA Testing

## Details

### DIFF
--- a/Runtime/CoreSettingsHandler.cs
+++ b/Runtime/CoreSettingsHandler.cs
@@ -15,6 +15,12 @@ namespace ReadyPlayerMe.Core
                 if (coreSettings == null)
                 {
                     coreSettings = Resources.Load<CoreSettings>(SETTINGS_PATH);
+#if UNITY_EDITOR
+                    if (coreSettings == null)
+                    {
+                        coreSettings = CreateCoreSettings();
+                    }
+#endif
                 }
                 return coreSettings;
             }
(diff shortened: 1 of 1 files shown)


HarrisonHough:
to test import sdk into a unity project and update to these URLs in the Packages/manifest.json
```
    "com.atteneder.gltfast": "https://github.com/atteneder/glTFast.git#v5.0.0",
    "com.readyplayerme.avatarloader": "https://github.com/readyplayerme/rpm-unity-sdk-avatar-loader.git#feature/core-settings-fix",
    "com.readyplayerme.core": "https://github.com/readyplayerme/rpm-unity-sdk-core.git#feature/core-settings-fix",
    "com.readyplayerme.webview": "https://github.com/readyplayerme/rpm-unity-sdk-webview.git#release/v0.2.0",
```

srcnalt:
Works great, can this whole script be moved into Editor folder? then we would not need to mark its sections editor only, I assume saving subdomain in there also is editor only function.

srcnalt (approved):

HarrisonHough:
> Works great, can this whole script be moved into Editor folder? then we would not need to mark its sections editor only, I assume saving subdomain in there also is editor only function.

Yep that's the plan it would just require some refactoring (as there a number of places where CoreSettings is accessed at runtime (outside of editor) so I'll make a separate task that can be done after this release.

PR #50, opened 2023-02-27
[3565717884] Added welcome window for Quick start + Analytics panel

HarrisonHough:
## Description

-   Added welcome window for quick start sample
- also merged the Analytics confirmation panel into this window
- If quick start button clicked it will import the sample and open the scene

#### Added

-   Welcome window for quick start and analytics popup
- quick start editor UI panel
- analytics editor UI panel

#### Removed

-  separate analytics window

## How to Test

## Checklist

-   [ ] Changelog is updated.

## QA Testing

## Details


HarrisonHough:
To test open a fresh or empty unity project and import this 
https://github.com/readyplayerme/rpm-unity-sdk-core.git#feature/quick-start-sample

The welcome window should popup (unless you have previously set don't ask again on the project)

image

After you click enable or disable analytics the window should show the QuickStart panel. 

image

Then click quickstart to open (or import then open) the quick start scene

Editor/UI/Components/QuickStartPanel.cs
```diff
@@ -0,0 +1,136 @@
...
+            EditorGUILayout.EndHorizontal();
+            GUILayout.Space(8);
+        }
+
+        private Sample GetQuickStartSample()
+        {
+            var samples = Sample.FindByPackage(LOADER_PACKAGE, null).ToArray();
+            var quickStartSample = samples.First(x => x.displayName == QUICKSTART_SAMPLE_NAME);
+            return quickStartSample;
+        }
```
rYuuk, line 121: oh nice, didn't knew about this.

rYuuk (approved):
🟢 Great Job, refactoring of editor window looks really nice. Just one query, I see that `IEditorWindowComponent` is only inherited but no referenced anywhere, is this intentional?
image

`PS: Don't forget to change module list to develop before merging. `

Also there seems to be no PR in avatar loader for quick sample?

HarrisonHough:
> 🟢 Great Job, refactoring of editor window looks really nice. Just one query, I see that `IEditorWindowComponent` is only inherited but not referenced anywhere, is this intentional? image
> 
Good point I guess it's kind of redundant. This interface was there already so I thought I would just use it as with the footer banner. 
I will clean-up the unused stuff tomorrow. 

Will make a PR tomorrow for the avatar loader (quick start sample)

srcnalt:
Importing into fresh Unity 2021 project and getting the error message below:
```UnityException: get_dataPath is not allowed to be called from a ScriptableObject constructor (or instance field initializer), call it in OnEnable instead. Called from ScriptableObject 'WelcomeWindow'.
See "Script Serialization" page in the Unity Manual for further details.
ReadyPlayerMe.Core.Editor.WelcomeWindow..cctor () (at Library/PackageCache/com.readyplayerme.core@36a4d56029/Editor/UI/EditorWindows/WelcomeWindow.cs:18)
Rethrow as TypeInitializationException: The type initializer for 'ReadyPlayerMe.Core.Editor.WelcomeWindow' threw an exception.```

File: Editor/UI/EditorWindows/WelcomeWindow.cs [hunk 2]
```diff
@@ -0,0 +1,118 @@
...
+    {
+        private Banner banner;
+        private GUIStyle buttonStyle;
+        private GUIStyle descriptionStyle;
+        private GUIStyle headingStyle;
+        private AnalyticsPanel analyticsPanel;
+        private QuickStartPanel quickStartPanel;
+        private bool displayQuickStart;
+
+        public static readonly string NeverAskAgainPref = $"{Application.dataPath}rpm-sdk-metrics-never-ask-again";
```
srcnalt, line 18: Can't read this property here.

srcnalt:
You need to bring back the editor component interface. Interfaces are not useless code pieces they are contracts that other classes comply with. Banner and Footer are classes of similar structure and should follow a certain rule, and if we create other UI elements similar to those they should also do the same - which will be more apparent when we move forward with optional modules with their own UI structure. The only time this will be deprecated is when we move forward with Unity UI Toolkit.

srcnalt (approved):

Editor/UI/EditorWindows/WelcomeWindow.cs
HarrisonHough: yep fixed now

PR #90, opened 2023-07-20
[SDK-259] change meshlod to lod

rk132:
## Dependencies
```Package
Avatar Loader: branch
```

## Description

-   pass lod parameter to the avatar loader query instead of meshlod

#### Updated

-   pass lod parameter to the avatar loader query instead of meshlod

## How to Test

- Open up test scene for avatar lod
- When changing distance, the LOD is also applied to the head.

## Checklist

-   [ ] Changelog is updated.

[SDK-259]: https://ready-player-me.atlassian.net/browse/SDK-259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/Editor/AvatarLoaderTests.cs
+++ b/Tests/Editor/AvatarLoaderTests.cs
@@ -13,7 +13,6 @@ namespace ReadyPlayerMe.Core.Tests
     public class AvatarLoaderTests
     {
         private GameObject avatar;
-        private GameObject avatar2;
 
         [TearDown]
         public void Cleanup()
@@ -24,10 +23,6 @@ namespace ReadyPlayerMe.Core.Tests
             if (avatar != null)
             {
                 Object.DestroyImmediate(avatar);
-            }
-            if (avatar2 != null)
-            {
-                Object.DestroyImmediate(avatar2);
             }
         }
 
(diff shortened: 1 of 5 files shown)


Tests/Editor/AvatarLoaderTests.cs
```diff
@@ -11,7 +13,8 @@ namespace ReadyPlayerMe.Core.Tests
     public class AvatarLoaderTests
     {
         private GameObject avatar;
-        
+        private GameObject avatar2;
```
HarrisonHough, line 16: This is actually not necessary, you can just re-use the same avatar object field, overriding it once you have the vertex count of the first loaded avatar.

HarrisonHough (changes requested):
Looks good! Just 1 small suggestion in the comments to simplify things a little.

Tests/Editor/AvatarLoaderTests.cs
```diff
@@ -173,5 +180,37 @@ public IEnumerator AvatarLoader_Cancel_Loading()
             Assert.AreNotEqual(FailureType.None, failureType);
             Assert.AreEqual(null, avatar);
         }
+
+        [UnityTest, CanBeNull]
```
HarrisonHough, line 184: Also forgot to mention, why was this CanBeNull macro added? 

I believe it's not necessary, and can be removed along with the 
```cs
using JetBrains.Annotations;
```
At the top

rYuuk (commented):
Please can you align the title of PR with our PR formats(you can check the closed PRs). Also for future, regarding branch name, we don't use ticket number in there as it feels redundant.

HarrisonHough (approved):
Looks good. 
Great stuff!

PR #98, opened 2023-08-02
[SDK-301] Add runtime analytics to quick start

rYuuk:
## Description

- Added scripts for runtime analytics
- Added new analytic events for quick start scene
- Restructure/renaming for few analytics files
- Analytics for Integration guide

## How to Test

- Enable analytics 
- Run quick start scene and load personal avatar
- Open integration guide and try few documentation links
- Observe events on amplitude

## Checklist

-   [ ] Changelog is updated.

[SDK-301]: https://ready-player-me.atlassian.net/browse/SDK-301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Runtime/Analytics/AmplitudeEventLogger.cs
```diff
@@ -0,0 +1,104 @@
...
+            var webRequestDispatcher = new WebRequestDispatcher();
+            var response = await webRequestDispatcher.SendRequest<Response>(url, HttpMethod.POST, headers, payload);
+
+            if (!response.IsSuccess)
+            {
+                throw new Exception(response.Error);
+            }
+        }
+        
+        private static string GetAnalyticsTarget()
```
HarrisonHough, line 94: An alternative to this function could be to just have a public static string declared up the top, something like this. Not really sure if it is better or worse though :D 

#if RPM_DEVELOPMENT
        public static readonly string TARGET_ENVIRONMENT = "unity";
#else
       public static readonly string TARGET_ENVIRONMENT = "unity-dev";
#endif

Runtime/Analytics/AmplitudeRuntimeLogger.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public void LogLoadPersonalAvatarButton()
+        {
+            LogEvent(PERSONAL_AVATAR_LOAD_BUTTON_CLICKED);
+        }
+
+        private void LogEvent(string eventName, Dictionary<string, object> eventProperties = null, Dictionary<string, object> userProperties = null)
+        {
+            if (!CoreSettingsHandler.CoreSettings.EnableAnalytics) return;
```
HarrisonHough, line 35: we could potentially move this check inside the AmplitudeEventLogger.LogEvent as we do the same if return statement in the Editor logger. To remove a tiny amount of code duplication.

Runtime/Data/ApplicationData.cs
```diff
@@ -6,7 +6,7 @@ namespace ReadyPlayerMe.Core
 {
     public static class ApplicationData
     {
-        private const string SDK_VERSION = "v2.0.0";
+        private const string SDK_VERSION = "v3.0.0";
```
HarrisonHough, line 9: oops I totally forgot about this variable :D

HarrisonHough (approved):
Looks good. Some small things mentioned in the comments, no blockers for me though. 
Events seem to work too,

Runtime/Analytics/AmplitudeRuntimeLogger.cs
rYuuk: I agree it could remove duplication, but this will make the AmplitudeEventLogger script dependent on CoreSettings. This script is currently decoupled and standalone with very few dependency.

PR #102, opened 2023-08-11
[SDK-332] Hotfix/v3.1.1

HarrisonHough:
## Description

-   fix for analytics going to wrong environment 

## How to Test

## Checklist

-   [ ] Changelog is updated.

### DIFF
--- a/Editor/Module Management/ModuleInstaller.cs
+++ b/Editor/Module Management/ModuleInstaller.cs
@@ -22,7 +22,7 @@ namespace ReadyPlayerMe.Core.Editor
 
         private const int THREAD_SLEEP_TIME = 100;
         private const string PROGRESS_BAR_TITLE = "Ready Player Me";
-        private const string RPM_DEVELOPMENT_SCRIPTING_SYMBOL = "RPM_DEVELOPMENT";
+        private const string READY_PLAYER_ME_SYMBOL = "READY_PLAYER_ME";
         private const string GLTFAST_SYMBOL = "GLTFAST";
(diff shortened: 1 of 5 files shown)


Editor/Module Management/ModuleInstaller.cs
```diff
@@ -35,19 +35,25 @@ public static class ModuleInstaller
...
 
+
         /// <summary>
         ///     Called when a package is added, removed or changed.
         /// </summary>
         /// <param name="args">Describes the <c>PackageInfo</c> entries of packages that have just been registered.</param>
         private static void OnRegisteredPackages(PackageRegistrationEventArgs args)
         {
             Events.registeredPackages -= OnRegisteredPackages;
+#if RPM_DEVELOPMENT
```
MaxAndreassenRPM, line 54: What determines this being set to true? Is it effectively an env var based on the build?

MaxAndreassenRPM (approved):

rYuuk (approved):
🟢 Thanks for making the fix so fast.

Editor/Module Management/ModuleInstaller.cs
HarrisonHough: Well kinda. 
It's a scripting define symbol that can be added from the project settings. But by default it won't be there. 
Currently it serves the purpose of :
- preventing the auto installer from running (used for Unity-SDK or anybody wanting to install the package into Assets folder for development purposes)
- and of course it also determines which environment the analytics gets sent to, more so for us internally so we don't add lots of garbage events during development 


image

PR #104, opened 2023-08-23
[SDK-321] Feature/gtlfast symbol fix

HarrisonHough:
## Description

-   This PR adds a fix for handling an issue where GLTFAST symbol is not recognized after being added. 
Please test on a new FRESH project by importing from this URL 
`https://github.com/readyplayerme/rpm-unity-sdk-core.git#feature/gtlfast-symbol-fix`

## How to Test

-   create new Unity project with unique project name and import package from this branch 
`https://github.com/readyplayerme/rpm-unity-sdk-core.git#feature/gtlfast-symbol-fix`
Then try to immediately load an avatar from Avatar Loader window
image

## Checklist

-   [ ] Changelog is updated.

[SDK-321]: https://ready-player-me.atlassian.net/browse/SDK-321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Editor/Module Management/ModuleInstaller.cs
+++ b/Editor/Module Management/ModuleInstaller.cs
@@ -189,8 +189,7 @@ namespace ReadyPlayerMe.Core.Editor
             SetDefineSymbol(BuildTargetGroup.WebGL);
             SetDefineSymbol(BuildTargetGroup.Android);
             SetDefineSymbol(BuildTargetGroup.iOS);
-            CompilationPipeline.RequestScriptCompilation();
-            CodeEditor.CurrentEditor.SyncAll();
+            BuildPipeline.BuildPlayer(new BuildPlayerOptions());
         }
 
         private static void SetDefineSymbol(BuildTargetGroup target)
--- a/Editor/Module Management/ModuleInstaller.cs
+++ b/Editor/Module Management/ModuleInstaller.cs
@@ -39,6 +39,9 @@ namespace ReadyPlayerMe.Core.Editor
 
         static ModuleInstaller()
         {
+    #if !GLTFAST
+            AddGltfastSymbol();
+    #endif
             Events.registeredPackages += OnRegisteredPackages;
             Events.registeringPackages += OnRegisteringPackages;
         }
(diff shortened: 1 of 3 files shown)


Editor/Module Management/ModuleInstaller.cs
```diff
@@ -178,24 +179,37 @@ public static PackageInfo[] GetPackageList()
...
+            foreach (BuildTarget target in Enum.GetValues(typeof(BuildTarget)))
+            {
+                BuildTargetGroup group = BuildPipeline.GetBuildTargetGroup(target);
+
+                if (group == BuildTargetGroup.Unknown)
+                {
+                    continue;
+                }
+
+                List<string> defineSymbols = PlayerSettings.GetScriptingDefineSymbolsForGroup(group).Split(';').Select(d => d.Trim()).ToList();
```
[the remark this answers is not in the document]
rYuuk: I am also for var as if the type is not evident you don't need to even hover rider adds it front of variable. Also we shouldn't care about people using notepad for coding 😅

Editor/UI/EditorWindows/AvatarLoaderEditor/AvatarLoaderEditor.cs
```diff
@@ -28,6 +28,9 @@ public class AvatarLoaderEditor : EditorWindow
         [MenuItem("Ready Player Me/Avatar Loader", priority = 0)]
         public static void ShowWindow()
         {
+#if !GLTFAST
+            ModuleInstaller.AddGltfastSymbol();
+#endif
```
rYuuk, line 33: This feels super hacky, what if someone doesn't open this window at all?

MaxAndreassenRPM:
FYI, I've just tested it and it worked with the avatar being loaded 🥳

Editor/UI/EditorWindows/AvatarLoaderEditor/AvatarLoaderEditor.cs
HarrisonHough: So the only case where a recompile wont happen before an avatar is loaded is if they open the Avatar loader Window. 
For example: 
- If they want to run a sample, they need to import the sample (which will trigger a recompile). 
- if the restart the editor, that will also trigger a recompile
- if they add a script to start using avatar loader it will trigger a recompile. 
BUT if they open avatar loader window and try load from there it would not, so the newly added symbol won't be recognized unless I do the above ^ 

But to your point I agree it is not the best way :D More than open to ideas though.

rYuuk: Makes perfect sense, i don't get why `CompilationPipeline.RequestScriptCompilation` doesn't work. But anyway this should be fine.

rYuuk (approved):
🟢 Works fine. Thanks for making this.

PR #110, opened 2023-08-28
[SDK-381] add test for avatar loader window

rk132:
## Description

-   add test to verify, that the prefab created is a directly linked to object in world without overrides or changes.

## How to Test

## Checklist

-   [ ] Changelog is updated.

[SDK-381]: https://ready-player-me.atlassian.net/browse/SDK-381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Editor/Utils/EditorUtilities.cs
+++ b/Editor/Utils/EditorUtilities.cs
@@ -40,8 +40,8 @@ namespace ReadyPlayerMe.Core.Editor
             return !string.IsNullOrEmpty(urlString) &&
                    (Regex.Match(urlString, SHORT_CODE_REGEX).Length > 0 || Uri.IsWellFormedUriString(urlString, UriKind.Absolute) && urlString.EndsWith(".glb"));
         }
-        
-        public static GameObject CreateAvatarPrefab(AvatarMetadata avatarMetadata, string path)
+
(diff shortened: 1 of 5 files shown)


HarrisonHough (commented):
This is actually a super interesting way of testing it. 
I realize now that I should have put more specific information in this task, I mostly didn't as I assumed it would go to me. 
But in any case, you could probably simplify this a lot and just test this function. 

image

Impementation would be like: 
1. Load avatar
2. Run prefab creation function on avatar
3. Load prefab from resources
3. Check all skinmesh != null 
@rYuuk what do you think?

HarrisonHough (changes requested):
Okay so I think that the Avatar Loader window test is good, but I think you should still also add a test specifically for the CreateAvatarPrefab function. 

The test would basically need to do this 
1. Load avatar
2. Run prefab creation function on avatar
3. Load prefab from resources
4. Check all skinmesh != null

HarrisonHough (approved):
I was gonna ask why the extra param was added, but then I realized it was because of the test :D 
Nice work!

PR #112, opened 2023-08-29
Hotfix/v3.2.2

HarrisonHough:
### Fixed
- glTFast symbol check ahead of package import removed.
- glTFast class type check to prevent define symbol add is removed.
- Settings asset creation in class with `InitializeOnLoad` attribute, refactored to create asset after asset indexing, using delayCall.

### DIFF
--- a/.github/latest.md
+++ b/.github/latest.md
@@ -2,5 +2,5 @@
 ## Changelog
 
 ### Fixed
-- issue of missing mesh references when prefabs were created by avatar loader window
+- issue where order of operations caused scriptable object creation errors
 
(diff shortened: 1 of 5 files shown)


srcnalt:
#### Notes:
- There was a race condition in module loading and adding the define symbols. The order of ops should be Core import, symbol addition then glTfast import. Symbol add can also be placed last, but since thread is blocked due to package import blocked code sections in Core won't cause trouble.

- Unable to import newly created asset error was due to attempt of creation asset in a class marked InitializeOnLoad. Unity does not suggest creating assets before they are indexed, issue solved by delegating the action to EditorApplication.delayCall which is run after asset indexing is completed.

PR #121, opened 2023-09-08
[SDK-425] feat: display empty subdomain on quick start

HarrisonHough:
## Description

- this PR includes changes to the setup guide window. It will now default to an empty subdomain field and prevent clicking next unless they enter subdomain OR click the "I don't have an account..." toggle instead
- if users close this window without doing anything it will be in the same state next open (still empty)
- conversely if you do enable the "I don't have an account.." toggle it will remember that next time you open the window. 

image

## How to Test

-   Import the branch in a new project and observe the setup guide popup
- Subdomain field should be empty and display error icon next to it
- next button should be disabled (preventing progression to next screen)
- the "I don't have an account..." toggle should be unchecked by default
- If you close the window without doing anything else it should be in the same state if you reopen it

## Checklist

-   [ ] Changelog is updated.

[SDK-425]: https://ready-player-me.atlassian.net/browse/SDK-425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-425]: https://ready-player-me.atlassian.net/browse/SDK-425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Editor/UI/EditorWindows/SetupGuide/SetupGuide.cs
+++ b/Editor/UI/EditorWindows/SetupGuide/SetupGuide.cs
@@ -46,13 +46,12 @@ namespace ReadyPlayerMe.Core.Editor
         public void CreateGUI()
         {
             visualTreeAsset.CloneTree(rootVisualElement);
+            InitializeFooter();
             panel = new[]
             {
                 InitializeSubdomainPanel(),
                 InitializeAnalyticsPanel()
             };
-
-            InitializeFooter();
             StartStateMachine();
         }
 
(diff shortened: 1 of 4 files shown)


Editor/UI/EditorWindows/SetupGuide/SetupGuide.cs
```diff
@@ -74,18 +74,29 @@ private VisualElement InitializeSubdomainPanel()
             {
                 nextButton.SetEnabled(!string.IsNullOrEmpty(subdomain));
             };
-
-            subdomainPanel.Q<Toggle>(USE_DEMO_SUBDOMAIN_TOGGLE).RegisterValueChangedCallback(x =>
+            if (!ProjectPrefs.GetBool(USE_DEMO_SUBDOMAIN_TOGGLE) && CoreSettingsHandler.CoreSettings.Subdomain == CoreSettings.DEFAULT_SUBDOMAIN)
```
rk132, line 77: Wouldn't this break the flow for the users, if they open the Setup Guide and close it? Could we do something like this, that first time, the CoreSettings.Subdomain is set to empty string and we can check that it isn't empty. So for the users, who currently have the CoreSettings.Subdomain set to demo, it would stay as demo?

rk132: That way, we basically don't need the USE_DEMO_SUBDOMAIN_TOGGLE

rk132: Ahh, just realised, that you need to know somehow, that the subdomain toggle is switched on.

rk132 (approved):
Looks good from my side

Editor/UI/EditorWindows/SetupGuide/SetupGuide.cs
HarrisonHough: Yep admittedly it is overly complicated logic. But it is done that way to match the product UX requirements discussed with Bernhard. Basically, this logic enforces the blocking of the next step if they haven't changed from demo && demo checkbox NOT checked.

PR #124, opened 2023-09-12
[SDK-454] feature/settings-handler-refactor

HarrisonHough:
## Description

- refactored core settings handler to remove all editor logic
- also includes some other edits required to make this work
- added error logging for case where settings asset does not exist. 

## How to Test

-  install or check out this branch
- please check everything related to core settings loading, accessing, editing
- this includes changing subdomain, analytics and other settings from our various editor windows
- what happens if you delete the scriptable object does it break in editor or for runtime? 
- try building a sample scene and running, are there build or runtime errors? (related to core settings not loading or something) 
- when installing fresh does core settings get created as expected? 

## Checklist

-   [ ] Changelog is updated.

[SDK-454]: https://ready-player-me.atlassian.net/browse/SDK-454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Editor/Analytics/AnalyticsEditorLogger.cs
```diff
@@ -16,16 +18,14 @@ public static void Enable()
         {
             IsEnabled = true;
```
MaxAndreassenRPM, line 19: I know it's not in your PR, but you could simplify this by removing the local IsEnable variable manual setting, and then on line 15 just make it a computed property... If Unity supports c# computed properties 😅 

`public static bool IsEnabled => CoreSettingsHandler.CoreSettings.EnableAnalytics;`

Runtime/CoreSettingsHandler.cs
```diff
@@ -1,71 +1,24 @@
 ﻿using ReadyPlayerMe.Core.Data;
-using UnityEditor;
 using UnityEngine;
 
 namespace ReadyPlayerMe.Core
 {
     public static class CoreSettingsHandler
```
MaxAndreassenRPM, line 6: I'd rename this to `CoreSettingsLoader` as that's the only thing it does.

Editor/Analytics/AnalyticsEditorLogger.cs
HarrisonHough: right good idea i guess it should work, just not sure if it will update or it only gets the value initially 🤔 
Will test it out

Runtime/CoreSettingsHandler.cs
MaxAndreassenRPM: It still sort of loads it though, except if it's already loaded then it has it to hand so can provide straight away. From the point of view of a piece of consuming code then load would make sense to me. Maybe `CoreSettingsProvider` could work? It just needs to be something that is not as generic as handler (because that means nothing 😅), and instead is something that implies that it's the thing which gives you the core settings (e.g. loader, provider, fetcher etc.)

HarrisonHough: right that makes sense. 
One other thing that also makes me a little hesitant (that I just realized) changing the name of this class in general because  multiple modules/packages use this class (CAC and WebView) so changing this will also require people to update these packages. 
The WebView is not so much an issue because we will be moving the core/shared functionality into core (as per another task).

HarrisonHough: probably if we change the name we should consider it a breaking change. We can also include better handling of breaking changes in this release then?

MaxAndreassenRPM: Ah, that's annoying. Is it worth making this change a breaking change which would then need a whole major version bump 🤔 - probably not to be honest 😅 - Maybe we could include renaming this the next time we do have to do a breaking change?

HarrisonHough: Well actually, if we improve our auto updater in the release, we can force it to install the latest of other packages (as required) if they choose to upgrade core. if that makes sense. 
Then we wouldn't need to make it a breaking change. 
However if developers have other scripts somewhere outside the packages that use CoreSettingsHandler then it would still cause issues. 
Perhaps just including in docs or something might be enough

Editor/Analytics/AnalyticsEditorLogger.cs
HarrisonHough: actually one potential issue I see with this ^ 
Is that order of operations might cause problems when initially installing the package. 
As we create the CoreSettings scriptable object from C# after the package is imported. I assume that this variable would be computed before the SO (scriptable object) is created so it would return null when trying to load it.

Runtime/CoreSettingsHandler.cs
HarrisonHough: I made the change but I just realized that it also makes it difficult to test first time import due to the name change 
image

HarrisonHough: I reverted the class rename. 
If we want to go with it we can do it in a separate PR and probably also include docs or installer updates to try and minimize install errors that this change will cause

Editor/Analytics/AnalyticsEditorLogger.cs
HarrisonHough: So it works if core settings already exists. 
But on first time install it tries to access the core settings (due to this code change) before core settings is created which causes an error. So I will keep the current logic with the extra variable.

MaxAndreassenRPM: Ah, that sucks, but fair enough 👍

MaxAndreassenRPM (approved):

rYuuk (approved):
🟢

Editor/Utils/CoreSettingsSetter.cs
```diff
@@ -0,0 +1,39 @@
+using UnityEditor;
+
+namespace ReadyPlayerMe.Core.Editor
+{
+
```
rYuuk, line 5: space 👾

Runtime/Data/CoreSettings.cs
```diff
@@ -4,11 +4,17 @@ namespace ReadyPlayerMe.Core.Data
...
 
         public string Subdomain = DEFAULT_SUBDOMAIN;
         public string AppId = string.Empty;
         public bool EnableLogging;
         public bool EnableAnalytics;
+
+        public static CoreSettings Load()
+        {
+            return Resources.Load<CoreSettings>(RESOURCE_PATH);
+        }
```
rYuuk, line 18: What do you think of moving this to handler, so that this script is just the data?

rYuuk (approved):
🟢  Thanks for separating out different responsibility, now it satisfy SRP more. Please also have a look at minor comments.

MaxAndreassenRPM (approved):

Runtime/Data/CoreSettings.cs
HarrisonHough: sure can do that. 
to me it's a little more convenient to be able to go CoreSettings.Load() as a first-time developer wanting to know how I can get the core settings. Instead of having to know there is a separate CoreSettingsHandler class for loading. 
I understand though that CoreSettings should be just a data object. 
The other thing is that in this sense the handler should be renamed something like CoreSettingsRuntimeLoader or something and the name change will cause a breaking change. 

in any case I will move it back to handler :D

HarrisonHough: updated. will merge

Editor/Utils/CoreSettingsSetter.cs
HarrisonHough: cleaned

PR #126, opened 2023-09-27
[SDK-511] Hotfix/v3.2.4

HarrisonHough:
## Description

-   this PR includes a fix that will make WebView only auto-install on first import, if it is later remove it won't be automatically reimported. 

## How to Test

## Checklist

-   [ ] Changelog is updated.

[SDK-511]: https://ready-player-me.atlassian.net/browse/SDK-511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Editor/Module Management/ModuleInstaller.cs
+++ b/Editor/Module Management/ModuleInstaller.cs
@@ -23,6 +23,9 @@ namespace ReadyPlayerMe.Core.Editor
         private const int THREAD_SLEEP_TIME = 100;
         private const string PROGRESS_BAR_TITLE = "Ready Player Me";
         private const string GLTFAST_SYMBOL = "GLTFAST";
+        private const string READY_PLAYER_ME_SYMBOL = "READY_PLAYER_ME";
+        private const string GLTFAST_NAME = "com.atteneder.gltfast";
+        private const string WEBVIEW_NAME = "webview";
 
         private const string MODULE_INSTALLATION_SUCCESS_MESSAGE =
             "All the modules are installed successfully. Ready Player Me avatar system is ready to use.";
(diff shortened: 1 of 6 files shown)


Editor/Module Management/ModuleInstaller.cs
```diff
@@ -101,12 +100,6 @@ public static void AddModuleRequest(string identifier)
         {
             var startTime = Time.realtimeSinceStartup;
             AddRequest addRequest = Client.Add(identifier);
-
-            // reset first time setup flag to ensure setup guide window is displayed
-            if (identifier.Contains(AVATAR_LOADER_SUBSTRING))
-            {
-                ProjectPrefs.SetBool(ProjectPrefs.FIRST_TIME_SETUP_DONE, false);
-            }
```
rk132, line 109: When was this reset or why was it necessary?

rk132 (approved):
Looks good

One small question

Editor/Module Management/ModuleInstaller.cs
HarrisonHough: Tbh I don't remember why exactly, but it was used for the separate AvatarLoader package, and since that's no longer required (due to 3.0 core release) I removed.

rYuuk (changes requested):
🔴 Seems to be not working 

https://github.com/readyplayerme/rpm-unity-sdk-core/assets/1121080/3837bea1-9635-4412-bd90-8aac1da499da

Editor/Module Management/ModuleInstaller.cs
```diff
@@ -127,9 +120,12 @@ public static void AddModuleRequest(string identifier)
         private static ModuleInfo[] GetMissingModuleNames()
         {
             PackageInfo[] installed = GetPackageList();
-            IEnumerable<ModuleInfo> missing = ModuleList.Modules.Where(m => installed.All(i => m.name != i.name));
-
-            return missing.ToArray();
+            var missingModules = ModuleList.Modules.Where(m => installed.All(i => m.name != i.name)).ToList();
+            if (ProjectPrefs.GetBool(ProjectPrefs.FIRST_TIME_SETUP_DONE))
```
HarrisonHough, line 124: @rYuuk Any ideas on a better way to do this with a hotfix? 
Otherwise, I am somewhat tempted to just remove WebView from the installer list altogether as it would be the simplest approach that wouldn't cause errors other than lack of WebView by default :D

PR #135, opened 2023-10-20
[SDK-562] Merge avatar creator into core

rYuuk:
## Description

- Merged avatar creator into core
- Renamed avatar creator sample to sample creator

## How to Test

- Try sample creator 

## Checklist

-   [ ] Changelog is updated.

[SDK-562]: https://ready-player-me.atlassian.net/browse/SDK-562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -63,6 +63,11 @@
       "displayName": "WebGLExample",
       "description": "This sample includes a WebGL template and demonstrates how run the Ready Player Me Web avatar creator inside an iFrame for WebGL applications.",
       "path": "Samples~/WebGLExample"
+    },
+    {
+      "displayName": "SampleCreator",
```
HarrisonHough, line 68: So I know This name was my idea, but now I'm thinking that a better one would be more like. 
`AvatarCreatorExample` or `CreatorExample`
For 2 reasons. 
1. We already have a WebGLExample so it will be more consistent naming
2. But aside from that I think the AvatarCreator part should actually be first. 

Another thing that maybe we should consider (and I'm not sure the best approach) but we will have multiple examples in future EG VR, Mobile and perhaps whatever else. So we could ether have separate, specific names for them EG CreatorVRExample, CreatorMobileExample or we could potentially just have a AvatarCreatorExamples package that includes all of them. Devs could later delete the ones they don't need? 
I will actually make a discussion thread in slack for this.

HarrisonHough (approved):
It actually looks all good. But I think before we merge we should have a quick discussion as a team about sample naming and structure (in regards to how to handle multiple creator examples)

HarrisonHough (approved):
Looks good! awesome work 👏

PR #144, opened 2023-11-02
[SDK-376] Add mesh opt import

rYuuk:
## Description

-  Refactored AvatarConfigEditor with ui toolkit

## How to Test

- Try different avatar config options and load avatar in a new project

## Checklist

-   [ ] Changelog is updated.

[SDK-376]: https://ready-player-me.atlassian.net/browse/SDK-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HarrisonHough (approved):
Changes are good and the auto import seems to work. 
I would say that the UI layout for avatar config could be improved slightly, like lining up the compression checkboxes. And expanding the texture channel stuff so that there is no need to scroll (if possible). 
But these are minor and non-blocking for me. We could always make a separate low-priority task for this if we want. 
image

PR #145, opened 2023-11-06
[SDK-513] Add appid to setup guide

rYuuk:
## Description

- Added app id to setup guide

## How to Test

-  Import package and open setup guide

## Checklist

-   [ ] Changelog is updated.

[SDK-513]: https://ready-player-me.atlassian.net/browse/SDK-513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HarrisonHough (approved):
Seems to work, but is the appID supposed to have a red error icon next to it if it's not set? 
It might be weird that you block the next button but don't highlight the field with the problem like we do with subdomain
image

PR #146, opened 2023-11-07
SDK-589 - Adds package install event

MaxAndreassenRPM:
I tried to add a unit test, but the static nature of a lot of the Unity SDK made this very difficult. I think we should spend some time on making the SDK more testable soon, perhaps installing something like Moq so we can at least mock out some of the more awkward dependencies as a starting point to test stuff in isolation -> https://docs.unity3d.com/Packages/nuget.moq@2.0/manual/index.html 

I've also include a little bit of tidying up in the classes I've touched in this PR.

Ignore all the commits, I will squash upon merging 😅.

image


Editor/Analytics/AmplitudeEditorLogger.cs
```diff
@@ -262,24 +274,25 @@ private void GenerateSessionId()
...
+                { Properties.ALLOW_ANALYTICS, allow }
             });
         }
-        
-        private void LogEvent(string eventName, Dictionary<string, object> eventProperties = null, Dictionary<string, object> userProperties = null)
+
+        private void LogEvent(string eventName, Dictionary<string, object> eventProperties = null, Dictionary<string, object> userProperties = null, bool force = false)
         {
-            if (!isEnabled) return;
+            if (!isEnabled && !force) return;
```
rYuuk, line 294: We should not be doing this. If somebody have disabled analytics, we shouldn't send any events. 
@bfercher Please confirm whether this is correct.

Editor/Analytics/AmplitudeEditorLogger.cs
```diff
@@ -59,190 +61,200 @@ public void LogOpenProject()
...
+        }
+
+        public void LogPackageInstalled(PackageCoreInfo packageInfo, bool force = false)
+        {
+            LogEvent(EventName.INSTALL_PACKAGE, new Dictionary<string, object>
+            {
+                { "id", packageInfo.Id },
+                { "name", packageInfo.Name },
+                { "url", packageInfo.Url },
+            }, force: force);
```
rYuuk, line 244: I am not sure whether we should do this according to GDPR. Until and unless someone have enabled analytics we shouldn't send an event.

rYuuk (commented):
Good Job in figuring out unity package install cycle, but I think this need to be thought through in regards to compliance.

HarrisonHough (commented):
Not entirely sure why but when I test on a new project in 2022.3.1 on Wiundows the event fired 4 times 
image

Editor/Analytics/AmplitudeEditorLogger.cs
MaxAndreassenRPM: Yup, it's primary reason for existing is because you can't opt in to analytics before you install the package... so we have to skip the analytics check in this use-case if we want to capture the package install event information. We could do it a different way which doesn't introduce a 'force' boolean, but it would still be doing the same thing 😅

Editor/Analytics/AmplitudeEditorLogger.cs
MaxAndreassenRPM: It's not a GDPR concern since it's not in anyway personal data, but still, I understand the desire to discuss it, so let's do so tomorrow in stand-up.

MaxAndreassenRPM:
> Not entirely sure why but when I test on a new project in 2022.3.1 on Wiundows the event fired 4 times image

Interesting, is it replicatable?

Editor/Analytics/AmplitudeEditorLogger.cs
MaxAndreassenRPM: Found a way to achieve the same thing whilst waiting for consent from the dev 🙏

Editor/Analytics/AmplitudeEditorLogger.cs
MaxAndreassenRPM: We discussed in stand-up and found a way to achieve it without skipping the acceptance 🙏

HarrisonHough:
> > Not entirely sure why but when I test on a new project in 2022.3.1 on Wiundows the event fired 4 times image
> 
> Interesting, can you replicate?

yep seems to happen every time on my work laptop 

image

Editor/Analytics/AmplitudeEditorLogger.cs
```diff
@@ -59,190 +61,201 @@ public void LogOpenProject()
...
         public void LogAvatarCreatorSampleImported()
         {
-           LogEvent(Constants.EventName.AVATAR_CREATOR_SAMPLE_IMPORTED);
+            LogEvent(EventName.AVATAR_CREATOR_SAMPLE_IMPORTED);
         }
-        
+
+        public void LogPackageInstalled(PackageCoreInfo packageInfo)
+        {
+            LogEvent(EventName.INSTALL_PACKAGE, new Dictionary<string, object>
```
HarrisonHough, line 239: you could perhaps instead add a PackageInfo.ToDictionary function instead of needing to create a new one here? Only thing you might need to do is make sure of correct capitalization. Just an idea though :D

HarrisonHough (approved):
Generally it looks fine. But I do get the install event being called multiple times (4 to be exact) on my work laptop though. Haven't tried other machines yet though.

Editor/Analytics/AmplitudeEditorLogger.cs
MaxAndreassenRPM: Interesting idea! I sort of like it as it is, since it's explicit to the method and you can clear look here to see exactly what properties are being sent to Amplitude. Maybe worth exploring though 🤔

PR #149, opened 2023-11-13
Hotfix/v4.1.0

HarrisonHough:
## Description

-   fixing an issue that caused avatars to be stored locally even if caching was disabled. 

## How to Test

## Checklist

-   [ ] Changelog is updated.

### DIFF
--- a/Editor/UI/EditorWindows/AvatarLoaderEditor/AvatarLoaderEditor.cs
+++ b/Editor/UI/EditorWindows/AvatarLoaderEditor/AvatarLoaderEditor.cs
@@ -83,7 +83,6 @@ namespace ReadyPlayerMe.Core.Editor
                 avatarLoaderSettings = AvatarLoaderSettings.LoadSettings();
             }
             var avatarLoader = new AvatarObjectLoader();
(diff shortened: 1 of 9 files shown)


Editor/UI/EditorWindows/AvatarLoaderEditor/AvatarLoaderEditor.cs
```diff
@@ -119,6 +118,11 @@ private void Completed(object sender, CompletionEventArgs args)
             }
             var paramHash = AvatarCache.GetAvatarConfigurationHash(avatarLoaderSettings.AvatarConfig);
             var path = $"{DirectoryUtility.GetRelativeProjectPath(args.Avatar.name, paramHash)}/{args.Avatar.name}";
+            if (!AvatarLoaderSettings.LoadSettings().AvatarCachingEnabled)
+            {
+                Debug.LogWarning("Enable Avatar Caching to generate a prefab in the project folder.");
+                return;
+            }
```
rk132, line 125: This blocks it completely, so that it won't be created to the scene as well. Do we want it like that. Also should we use SDKLogger here

PR #150, opened 2023-11-13
Hotfix/v4.0.1

HarrisonHough:
## Description

-   a small fix for avatar caching enable/disable logic

## How to Test

## Checklist

-   [ ] Changelog is updated.

### DIFF
--- a/.github/latest.md
+++ b/.github/latest.md
@@ -2,4 +2,4 @@
 ## Changelog
 
 ### Breaking Changes
-- fixed an issue causing avatars to be stored locally even if caching was disabled by @harrisonhough in [#149](https://github.com/readyplayerme/rpm-unity-sdk-core/pull/149)
+- fixed an issue causing avatars to be stored locally even if caching was disabled by @harrisonhough in [#150](https://github.com/readyplayerme/rpm-unity-sdk-core/pull/150)
--- a/.github/latest.md
+++ b/.github/latest.md
@@ -2,11 +2,4 @@
 ## Changelog
 
 ### Breaking Changes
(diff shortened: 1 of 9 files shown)


Editor/UI/EditorWindows/AvatarLoaderEditor/AvatarLoaderEditor.cs
```diff
@@ -119,6 +118,11 @@ private void Completed(object sender, CompletionEventArgs args)
             }
             var paramHash = AvatarCache.GetAvatarConfigurationHash(avatarLoaderSettings.AvatarConfig);
             var path = $"{DirectoryUtility.GetRelativeProjectPath(args.Avatar.name, paramHash)}/{args.Avatar.name}";
+            if (!AvatarLoaderSettings.LoadSettings().AvatarCachingEnabled)
+            {
+                Debug.LogWarning("Enable Avatar Caching to generate a prefab in the project folder.");
+                return;
+            }
```
rk132, line 125: This blocks it completely, so that it won't be created to the scene as well. Do we want it like that. Also should we use SDKLogger here

HarrisonHough: oh but yes should use sdklogger! thanks will update

rk132 (approved):
Looks good from my side

.github/latest.md
```diff
@@ -2,11 +2,4 @@
 ## Changelog
 
 ### Breaking Changes
```
rYuuk, line 4: This shouldn't be a breaking change?

rYuuk (approved):
🟢 Tested avatar caching working fine now, a small comment.

.github/latest.md
HarrisonHough: oh good catch! will fix

PR #151, opened 2023-11-13
[] Small CAC refactoring in prep for CAC 2.0 and wizardless flow

HarrisonHough:
## Description

-   This PR adds some minor changes which should add some extra flexibility and help us in CAC 2.0.  Please share your feedback though and point out any potential floors in the updates. I will list some main changes and the reasoning below. 

**1. **Added an extra return property to the CreateAvatar(AvatarProperties avatarProperties) method (made it return a tuple same as CreateAvatarFromTemplate.** 
   **Reasoning:** To create an avatar from a photo with gender prediction enabled we need to also have the avatar properties, in particular the gender. By adding this extra property we can easily do that. 
**2. Removed gender from AvatarManager constructor**
    **Reasoning**: this makes AvatarManager easier to work with as you don't need to create a new one each time you change genders. Also CreateAvatar is passed avartarProperties (which can optionally have gender predefined) and CreateAvatarFromTemplate gets the gender from Template data. The only thing that may be useful in future is to add a separate "SetGender" function but at the moment I don't really see a use for it. 
**3. Removed partner parameter from CreateAvatarFromTemplate() function.** 
   **Reasoning:** Why would anybody ever want to pass a "partner" value that is not the same as in their CoreSettings ? Doing so would probably break API's anyways (due to non matching subdomain and API key
   
   

## How to Test

## Checklist

-   [ ] Changelog is updated.


rYuuk (approved):
🟢 Looks good

PR #157, opened 2023-11-20
[SDK-623] feat: update api to work with GLTFast 6.0

HarrisonHough:
## Description

-   This PR includes updates to support GLTF 6.0 it is a breaking change and it seems there isn't an elegant way to add or handle backwards compatibility. 

## How to Test

-  Importy gltfast 6.0 and our SDK (this branch) everything should work without errors

## Checklist

-   [ ] Changelog is updated.

[SDK-623]: https://ready-player-me.atlassian.net/browse/SDK-623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/GLTF/GltFastGameObjectInstantiator.cs
+++ b/Runtime/GLTF/GltFastGameObjectInstantiator.cs
@@ -40,6 +40,8 @@ namespace ReadyPlayerMe.Core
             {
                 // Use Node GameObject for first Primitive
                 meshGo = m_Nodes[nodeIndex];
+                // Ready Player Me - Parent mesh to Avatar root game object
+                meshGo.transform.SetParent(m_Parent.transform);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


Runtime/GLTF/GltFastGameObjectInstantiator.cs
```diff
@@ -41,8 +40,6 @@ public class GltFastGameObjectInstantiator : GameObjectInstantiator
             {
                 // Use Node GameObject for first Primitive
                 meshGo = m_Nodes[nodeIndex];
-                // Ready Player Me - Parent mesh to Avatar root game object
-                meshGo.transform.SetParent(m_Parent.transform);
```
rk132, line 45: do you know, why was the parent changed to the root game object?

rk132 (approved):
Looks good. One small question.

Runtime/GLTF/GltFastGameObjectInstantiator.cs
HarrisonHough: ah good point, let me check. Probably I need to add this back in

HarrisonHough: I added it back in. But when testing it out I couldn't notice any difference. Will check with robin once he is back

PR #158, opened 2023-11-20
[SDK-626] Feature/webgl helper update

HarrisonHough:
## Description

-   Reworks how the auto setup functionality works when importing webGL sample as WebGLHelper also must be in Assets/Plugins folder to work. 
- So this PR adds the auto moving of this folder
- It also moves the folder rather than copies. This is better as you don't end up with file duplicates and that not only bloats the project but causes issues with duplicate GUIDs 

## How to Test

-  import on new projet with this https://github.com/readyplayerme/rpm-unity-sdk-core.git#feature/webgl-helper-update and then import the WebGLSample and click yes and observe behavior 

## Checklist

-   [ ] Changelog is updated.

[SDK-626]: https://ready-player-me.atlassian.net/browse/SDK-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Samples~/WebGLSample/Editor/EditorWindows/SampleSetup.cs
+++ b/Samples~/WebGLSample/Editor/EditorWindows/SampleSetup.cs
@@ -61,8 +61,8 @@ namespace ReadyPlayerMe.Samples
                 Debug.LogWarning("Failed to find WebGLSample. No changes were done to project");
                 return;
             }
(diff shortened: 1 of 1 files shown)


rk132 (approved):
Looks good

PR #159, opened 2023-11-21
[SDK-597] feat: add gender select element

HarrisonHough:
## Description

-   In this PR is a simple Gender select element with minimal functionality and a gender selected event that includes the outfit gender parameter (which is used to request a male or female avatar 
- It also includes a simple prefab with 3 buttons and everything assigned and setup. 
image

## How to Test

## Checklist

-   [ ] Changelog is updated.

[SDK-597]: https://ready-player-me.atlassian.net/browse/SDK-597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Runtime/AvatarCreator/GenderSelectElement.cs
```diff
@@ -0,0 +1,45 @@
...
+        private void Awake()
+        {
+            InitializeButtons();
+        }
+
+        private void InitializeButtons()
+        {
+            if (maleButton != null)
+            {
+                maleButton.onClick.AddListener(() => OnGenderSelected?.Invoke(OutfitGender.Masculine));
```
MaxAndreassenRPM, line 30: Small thing, but could you not use the same nullable operator that you use in onDestroy and drop the if statements?

```suggestion
                maleButton?.onClick.AddListener(() => OnGenderSelected?.Invoke(OutfitGender.Masculine));
```

Runtime/AvatarCreator/GenderSelectElement.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            if (femaleButton != null)
+            {
+                femaleButton.onClick.AddListener(() => OnGenderSelected?.Invoke(OutfitGender.Feminine));
+            }
+        }
+
+        private void OnDestroy()
+        {
+            maleButton?.onClick.RemoveListener(() => OnGenderSelected?.Invoke(OutfitGender.Masculine));
```
MaxAndreassenRPM, line 41: Does this actually work to remove the listener? As they're actually both separate anonymous methods, so technically you're removing nothing? Unless Unity has some magic to match them up?

Would it work better to not have these as anonymous methods, and then add and remove the listener by reference to the method? Like suggested here: https://forum.unity.com/threads/how-do-i-remove-button-listeners-and-should-i.1318596/

```suggestion
            maleButton?.onClick.RemoveListener(SendMaleSelected);
            
            private void SendMaleSelected() {
               OnGenderSelected?.Invoke(OutfitGender.Masculine)
            }
```

MaxAndreassenRPM (commented):
A couple of comments / questions.

Runtime/AvatarCreator/GenderSelectElement.cs
HarrisonHough: It's a good point actually and I think you're right, it was a bit of copy pasta from chatgpt :D 
I will update to use a function just to be sure.

Runtime/AvatarCreator/GenderSelectElement.cs
HarrisonHough: Yep the reason I didn't do it that way is using '?.' on objects deriving from UnityEngine.Object might not correctly reflect if it is null.
Unity implements/overrides the '==' operator to make up for this but it doesn't work on ?.. 
image

Runtime/AvatarCreator/GenderSelectElement.cs
HarrisonHough: updated

Runtime/AvatarCreator/GenderSelectElement.cs
MaxAndreassenRPM: Ah, okay - annoying but makes sense ✅  - what do you think about changing the other one then so it's consistent? 🤔

MaxAndreassenRPM (approved):

PR #160, opened 2023-11-21
feat: added basic login element

HarrisonHough:
## Description

-   This PR adds a generic Login panel with all the required elements and functionality including a ready to use prefab with events on fail and success
- It also makes it's own request to send verification code and login with code

Screenshot 2023-11-21 084921

Screenshot 2023-11-21 084930

image

## How to Test

## Checklist

-   [ ] Changelog is updated.

### DIFF
--- a/Runtime/AvatarCreator/UI/LoginElement.cs
+++ b/Runtime/AvatarCreator/UI/LoginElement.cs
@@ -1,0 +1,57 @@
+using System;
+using ReadyPlayerMe.Core;
+using UnityEngine;
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+namespace ReadyPlayerMe.AvatarCreator
+{
+    public class LoginElement : MonoBehaviour
+    {
+        private const string TAG = nameof(LoginElement);
+        [Header("Input Fields")]
+        [SerializeField, Tooltip("Input field for entering email address")] private InputField emailField;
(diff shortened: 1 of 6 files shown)


Runtime/AvatarCreator/UI/LoginElement.cs
```diff
@@ -0,0 +1,57 @@
...
+            AuthManager.SendEmailCode(emailField.text);
+        }
+
+        public async void LoginWithCode()
+        {
+            try
+            {
+                if (await AuthManager.LoginWithCode(codeField.text))
+                {
+                    OnLoginSuccess?.Invoke();
```
MaxAndreassenRPM, line 41: Might it be a bit cleaner to keep consistency between the error/success approaches? Currently for 1 you have a separate method (error) and for the other you do it inline (success). I think either approach is fine since it's such a small thing, but I think consistency is the important thing here.

```suggestion
                    OnSignInSuccess();
                    
                    private void OnSignInSuccess() {
                       OnLoginSuccess?.Invoke();
                       SDKLogger.Log('Blah');
                    }
```

HarrisonHough: updated and renamed some stuff

MaxAndreassenRPM (approved):

PR #164, opened 2023-11-27
[SDK-600] Feature/avatar template element

HarrisonHough:
## Description

-   This PR includes a TemplateAvatar element that can be used to display avatar template icons and fire an event with the required data. It also includes a ButtonElement  class which should be usable in other cases. 
- it has I realize that we already have Asset and Avatar button classes in the sample but I wanted to keep this separate so as not to break the sample but also I think these simpler elements might be more re-usable. We can always update later as needed. 

## How to Test

-   easiest way to test is to checkout the **feature/template-avatar-element** branch in Unity-SDK repo and run the test scene in the TEMP folder. You should be able to:
1. run the scene
2. see the template buttons load
3. click a button
4. see selected icon
5. check console for print out on avatar ID and gender

image

## Checklist

-   [ ] Changelog is updated.

[SDK-600]: https://ready-player-me.atlassian.net/browse/SDK-600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Runtime/AvatarCreator/Data/AssetLibrary.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace ReadyPlayerMe.AvatarCreator
 {
-    public struct AssetData
+    public struct AssetLibrary
```
rYuuk, line 6: 💯

Runtime/AvatarCreator/Data/AssetLibrary.cs
```diff
@@ -17,9 +17,7 @@ public struct PartnerAsset
         [JsonConverter(typeof(GenderConverter))]
         public OutfitGender Gender;
         [JsonProperty("iconUrl")]
-        public string Icon;
-        [JsonProperty("maskUrl")]
-        public string Mask;
+        public string ImageUrl;
```
rYuuk, line 20: Maybe we should move this PartnerAsset to a separate script, what do you think?

rYuuk: Also I suppose this is breaking stuff in current avatar creator?

Runtime/AvatarCreator/TemplateFetcher.cs
```diff
@@ -7,38 +7,34 @@ namespace ReadyPlayerMe.AvatarCreator
...
+        }
 
-            while (!downloadRenderTasks.All(x => x.IsCompleted) && !ctx.IsCancellationRequested)
-            {
-                await Task.Yield();
-            }
+        public async Task FetchTemplateRendersParallel(List<TemplateData> templates)
+        {
+            var tasks = templates.Select(FetchTemplateRenderAsync);
+            await Task.WhenAll(tasks);
```
rYuuk, line 32: I am not sure if this is the right way, by opening so much api it confuses the usage of this class a bit. Though I understand why you did it. I have another proposition. So previously it was downloading all the templates json and downloading all the images parallelly but waiting for the download to complete and than returning. If you want to get the images as soon as they are downloaded, maybe you can add a callback(action) which is fired as soon as an image is downloaded.

Runtime/AvatarCreator/UI/Buttons/ButtonElement.cs
```diff
@@ -0,0 +1,29 @@
...
+        [SerializeField] private RawImage rawImage;
+
+        public void AddListener(Action action)
+        {
+            button.onClick.AddListener(action.Invoke);
+        }
+
+        private void OnDestroy()
+        {
+            button.onClick.RemoveAllListeners();
```
rYuuk, line 20: May this is not necessary but I am not sure.

Runtime/AvatarCreator/UI/Elements/TemplateAvatarElement.cs
```diff
@@ -0,0 +1,94 @@
...
+            if (avatarTemplateDataList == null || avatarTemplateDataList.Count == 0)
+            {
+                SDKLogger.LogWarning(TAG, "No templates found");
+            }
+        }
+
+        public async Task LoadTemplateRenders()
+        {
+            avatarTemplateDataList = await templateFetcher.GetTemplatesWithRenders();
+        }
```
rYuuk, line 50: A bit of over complication, this way I am not sure which function to call and in which order.

rYuuk (changes requested):
🟡 Nice work, I nit-picked a bit😅. Please let me know if the concerns are valid or not.

Runtime/AvatarCreator/Data/AssetLibrary.cs
HarrisonHough: So everything will (or should) still work. But test it out and let me know if it doesn't. The icons have been updated now so that they are already masked basically.

HarrisonHough: and yep we can move to a separate script

Runtime/AvatarCreator/UI/Elements/TemplateAvatarElement.cs
HarrisonHough: What would you suggest here instead? 
Perhaps I am going overboard with providing multiple [ublic functions, the main thing I was trying to enable was that devs could control how they want to do the loading without needing to create their own script. But maybe that isn't actually necessary. Because on the other hand, we kinda want devs to be creating their own scripts as it should (in theory) reduce the risk of issues related to package updates. 
It's kind of like utility vs simplicity 😅

Runtime/AvatarCreator/TemplateFetcher.cs
HarrisonHough: I see your point. I guess the main thing I am trying to do is not force devs to use it a particular way, enabling them to implement things however they like.

HarrisonHough: But I guess they can already do that with API requests to get just template data only 
`return await avatarAPIRequests.GetTemplates();`

Runtime/AvatarCreator/Data/AssetLibrary.cs
MaxAndreassenRPM: I am also generally in favour of having a script/file per object declaration, normally in a 'models' folder but not sure if that naming would be as common in the Unity world. We should stay consistent though, so we should commit to doing this everywhere, or nowhere IMO.

Also, regarding it being a breaking change, even if it is, it shouldn't be an issue as eventually this (and a re-creation of the avatar creator using these) will replace the existing one. However, I guess any breaking changes should be merged to a feature branch rather than develop for that thought process to work?

Runtime/AvatarCreator/TemplateFetcher.cs
MaxAndreassenRPM: I'm actually in favour of having the lower level layers (e.g. fetchers, or requests, or repositories as they'd be if we translate this to web) as open and flexible as possible, with the caveat that they stay within the bounds of their context entity so they don't get confusing (in this case 'templates'). The only thing I'd potentially change about this is a few small nitpicks that aren't crucial:

1. I wouldn't suffix the third method with Parallel. Since it's the only way of doing it we offer, it doesn't add any extra value to the user, and it sort of over explains the inner workings of the method. If we one day wanted to offer a way of linearly downloading all the template renders too (which would be a bit weird, but maybe 🤷‍♂️) then we use a strategy pattern and you'd pass in an enum of DownloadStrategy.Parallel, or DownloadStrategy.Linear. For now however, if it was me, I'd just had the method as `FetchTemplateRendersAsync`.

2. **_Maybe_** I'd split the template renders (bottom two methods) into a separate service / fetcher since the context is slightly different. However, I'm not as sure on this at the moment myself so just throwing it out there.

3. I thought we spoke about how GetTemplateAvatarImage wasn't really needed as all it really is in a image downloader? If you split this out into a generic image downloader then I think it'd become clearer how you might not even need this whole 'fetch template render' concept in here, since it's just fetching images...

4. I also like the prospect of a callback that's fired when a long running async thing is finished, but not sure how much value it would add here in this scenario 🤔

Runtime/AvatarCreator/UI/Elements/TemplateAvatarElement.cs
MaxAndreassenRPM: I guess you shouldn't know which function to call and in what order, it's up to you based on how you're building your experience as a partner. The purpose of our SDK is to offer utility, not to dictate the steps in which you call our API to build your game / experience.

`Because on the other hand, we kinda want devs to be creating their own scripts as it should (in theory) reduce the risk of issues related to package updates.`

We definitely want devs creating their own scripts, but this lends to having how it is currently, because those scripts would call these underlying SDK scripts. I guess the point isn't about whether they interact with the SDK through their own scripts or through the editor, but more about us offering utility that they can use in whichever means they choose. The minute a dev has to create a script to rework something in one of our lower level scripts then we've done something wrong.

Runtime/AvatarCreator/UI/Buttons/ButtonElement.cs
```diff
@@ -0,0 +1,37 @@
...
+using UnityEngine.UI;
+
+namespace ReadyPlayerMe.AvatarCreator
+{
+    /// <summary>
+    /// A basic button element with an icon and functionality to bind an action to the
+    /// buttons onClick event.
+    /// </summary>
+    [RequireComponent(typeof(Button))]
+    public class ButtonElement : MonoBehaviour
```
MaxAndreassenRPM, line 12: Is this component even needed? Why can't the avatar creator just expose a public variable to define a button prefab? The only thing this component does is enforce that the icon texture is set on the button, but even that is a bit too opinionated IMO. What if I want to go GTA style and have my buttons be the asset names, and not the icons?

image

HarrisonHough: valid point. I think that almost always when dealing with 2D UI you will have a Button component/element because generally to make something selectable or rather clickable it needs or be some sort of button. 
But you are right though they may want to have this component on some parent object or something and then be able to assign the button to something else in the inspector. 
The RequireComponent is sort of useless here anyways as I didn't find/assign it on start.

HarrisonHough: To support text buttons we would probably need to also add a function to Set Text and also add a Text element 🤔

Runtime/AvatarCreator/UI/Buttons/ButtonElement.cs
HarrisonHough: Yeah I'm pretty sure the it's super unlikely that this is necessary it was more of a good practice thing

HarrisonHough: cleaned

Runtime/AvatarCreator/UI/Buttons/ButtonElement.cs
HarrisonHough: leaving as is for now. If there is a need we can rework as needed based on feedback from devs

MaxAndreassenRPM (approved):

rYuuk (approved):
🟢  Thanks for making the changes.

PR #166, opened 2023-11-27
[SDK-607] Add selfie element

rYuuk:
## Description

- Added selfie element
- Clean up elements prefab and scripts

## How to Test

-  Put prefab in empty scene and also it will require login anonymously

## Checklist

-   [ ] Changelog is updated.


Runtime/AvatarCreator/UI/SelfieElement.cs
```diff
@@ -0,0 +1,55 @@
...
+            avatarProperties.Gender = gender;
+            avatarProperties.Assets = new Dictionary<Category, object>();
+
+            if (!AuthManager.IsSignedIn && !AuthManager.IsSignedInAnonymously)
+            {
+                SDKLogger.Log(TAG, "Not signed in");
+                return;
+            }
+            
+            var avatarManager = new AvatarManager(bodyType);
```
HarrisonHough, line 49: Minor thing but I just realized that bodyType is being set in multiple places. I asume that the setting in the constructor will override the AvatarProps one. But probably that may be confusing to devs.
Perhaps we should just remove bodyType from the constructor, what do you think?

rYuuk: Yes we could do that, but than we have to provide almost all of the functions of AvatarManager with bodyType.

Runtime/AvatarCreator/UI/ImageConfirmationElement.cs
```diff
@@ -0,0 +1,25 @@
+using UnityEngine;
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+namespace ReadyPlayerMe.AvatarCreator
+{
+    public class ImageConfirmationElement : MonoBehaviour
```
MaxAndreassenRPM, line 7: Nice job cleaning these up to all be in the same directory ✅

Runtime/AvatarCreator/UI/SelfieElement.cs
```diff
@@ -0,0 +1,55 @@
...
+        [SerializeField] private BodyType bodyType;
+        [SerializeField] private OutfitGender gender;
+
+        [Space(5)]
+        [Header("Events")]
+        public UnityEvent<GameObject,AvatarProperties> onAvatarCreated;
+
+        private void OnEnable()
+        {
+            imageConfirmationElement.onImageConfirmed.AddListener(OnPhotoCaptured);
```
MaxAndreassenRPM, line 23: @HarrisonHough thoughts on this? We'd so far aimed to stay away from configuring the listeners in the code, as per you suggestion on a previous PR. Maybe we should do the same here too?

Runtime/AvatarCreator/UI/SelfieElement.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        private void OnDisable()
+        {
+            imageConfirmationElement.onImageConfirmed.RemoveListener(OnPhotoCaptured);
+        }
+
+        private async void OnPhotoCaptured(Texture2D texture)
```
HarrisonHough, line 31: I think we should remove the imageConfirmationElement reference (dependency) and have some public function like LoadAvatarFromSelfie (or something similar) here instead. Then the function can just be assigned in the prefab itself. 
By doing this they can use this selfie element in combination with not only our other elements but also their own scripts without having to use our ImageConfirmation or PhotoCapture element. 
 EG if they have a native camera/file picker they can just pass the image to Load from selfie etc.

Runtime/AvatarCreator/UI/SelfieElement.cs
HarrisonHough: yep good point I just mentioned similar thing in the comment below.

Runtime/AvatarCreator/UI/SelfieElement.cs
HarrisonHough: What do you mean exactly by having to provide all functions to AvatarManager?

HarrisonHough (changes requested):
Overall looks good but some suggestions and discussion points in the comments

Runtime/AvatarCreator/UI/SelfieElement.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (!AuthManager.IsSignedIn && !AuthManager.IsSignedInAnonymously)
+            {
+                SDKLogger.Log(TAG, "Not signed in");
+                return;
+            }
+            
+            var avatarManager = new AvatarManager(bodyType);
+            var avatar = await avatarManager.CreateAvatar(avatarProperties);
+
+            onAvatarCreated?.Invoke(avatar.Item1, avatar.Item2);
```
MaxAndreassenRPM, line 52: What is avatar item1 and item2? Looking at this code I would have no idea what this means without stepping into it more.

Runtime/AvatarCreator/UI/SelfieElement.cs
```diff
@@ -0,0 +1,55 @@
+using System;
```
MaxAndreassenRPM, line 1: Should this element be named 'SelfieToAvatarElement?', since it's so deeply tied to creating an avatar?

HarrisonHough: Yep ➕ to this, seems more descriptive.

Runtime/AvatarCreator/UI/SelfieElement.cs
rYuuk: Though consumer of this event will know what those 2 elements are  
```
public UnityEvent<GameObject,AvatarProperties> onAvatarCreated;
```

MaxAndreassenRPM: True, but still, I have no idea what gameObject is being passed, other than that it is called Item1. I think the code should be as self documenting as possible.

Although it looks like c# added a way to get better naming with tuples, so maybe there is a middle ground? https://learn.microsoft.com/en-us/archive/msdn-magazine/2015/december/essential-net-designing-csharp-7

rYuuk: oh nice didn't knew that, will change it accordingly.

Runtime/AvatarCreator/UI/PhotoCaptureElement.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            onPhotoCaptured?.Invoke(texture);
+        }
+
+        private void InitializeCamera()
+        {
+            var webCamDevice = GetWebCamDevice();
+
+            SetupPhotoBoothTexture(webCamDevice?.name);
+            StartCamera();
```
HarrisonHough, line 57: Actually not relevant to this PR, something I kinda missed in Max's PR. But perhaps we shouldn't force them to start the camera on awake (this InitializeCamera function is called on awake, it is also kinda a hidden "side effect" of initializing. I also wonder if it the automatic GetCamDeveice() on start might be annoying for some as it basically requests access regardless of whether the element is actually used? 🤔  Maybe I'm just overthinking this part though.

HarrisonHough (approved):
I see what you mean about the body type thing for avatar manager, kinda annoying we need to add to the function params. But I do think it's still better then having the multiple places to set body type. Maybe in future there might be a way to improve it even further somehow. 

I mention one small thing in the comments but it's not specific to the changes in this PR and is kinda minor. 

But overall great work ! And thanks for making the changes 🙏

Runtime/AvatarCreator/UI/PhotoCaptureElement.cs
MaxAndreassenRPM: I could see the positives for moving out the StartCamera() method to be invoked outside of awake (maybe in a from the editor itself in the element prefab), but then you'd maybe want the whole initialize method to not be on awake right?

Although since this is so easy to change between, maybe it's one of those ones where it's a case of us getting it out there and seeing what partners say. If we get reports that it is annoying then it's easy for us to change it and to respond to feedback.

HarrisonHough (approved):
Looks good. Cool stuff

PR #167, opened 2023-11-28
[SDK-595] Remove headers for icon web requests

rYuuk:
## Description

-  Removed auth and app Id headers for template icon and asset icon requests

## How to Test

- Run a WebGL build with avatar creator sample and check if there are no errors

## Checklist

-   [ ] Changelog is updated.

### DIFF
--- a/Runtime/Utils/WebRequestDispatcher.cs
+++ b/Runtime/Utils/WebRequestDispatcher.cs
@@ -31,6 +31,7 @@ namespace ReadyPlayerMe.Core
             Dictionary<string, string> headers = null,
             string payload = null,
             DownloadHandler downloadHandler = default,
+            bool sendAppIdInHeader = true,
             CancellationToken ctx = new CancellationToken()) where T : IResponse, new()
         {
             using var request = new UnityWebRequest();
@@ -46,7 +47,10 @@ namespace ReadyPlayerMe.Core
                 }
             }
 
-            request.SetRequestHeader(APP_ID, CoreSettingsHandler.CoreSettings.AppId);
+            if (sendAppIdInHeader)
+            {
+                request.SetRequestHeader(APP_ID, CoreSettingsHandler.CoreSettings.AppId);
+            }
 
             downloadHandler ??= new DownloadHandlerBuffer();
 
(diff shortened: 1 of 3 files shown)


Runtime/Utils/WebRequestDispatcher.cs
```diff
@@ -31,6 +31,7 @@ public class WebRequestDispatcher
             Dictionary<string, string> headers = null,
             string payload = null,
             DownloadHandler downloadHandler = default,
+            bool sendAppIdInHeader = true,
```
MaxAndreassenRPM, line 34: It's a quick way to solve the problem, but I'm not a fan of how we've now got a param specific to one RPM-bespoke header in the params for all requests that get made in the SDK, even if it is optional. It adds extra confusion where, one day, a partner will look at this and go 'what the hell is an app id header' because this is just a generic WebRequestDispatcher and now it's got some non generic web request logic in it...

Would you be open to considering another approach? Maybe you could wrap this class in ReadyPlayerMeWebRequestDispatcher which handles the logic of passing down the AppID header in the already existing 'headers' param of this class. Then you'd use the ReadyPlayerMeWebRequestDispatcher across the project, but the image downloading would use the base WebRequestDispatcher? You'd then have better separation of concerns but achieve the same goal.

MaxAndreassenRPM (changes requested):

Runtime/Utils/WebRequestDispatcher.cs
rYuuk: I came up with the different and much more simpler approach. Please have a look.

Runtime/Utils/WebRequestDispatcherExtension.cs
```diff
@@ -74,10 +38,7 @@ public static class WebRequestDispatcherExtension
             var headers = new Dictionary<string, string>();
             if (!url.Contains(CLOUDFRONT_IDENTIFIER)) // Required to prevent CORS errors in WebGL
             {
-                foreach (KeyValuePair<string, string> header in CommonHeaders.GetRequestHeaders())
-                {
-                    headers.Add(header.Key, header.Value);
-                }
+                headers = CommonHeaders.GetRequestHeaders();
```
MaxAndreassenRPM, line 41: 😍

PR #171, opened 2023-11-29
SDK-669 - environment file

MaxAndreassenRPM:
I really hate this approach as I couldn't get the env file variable declaration to be simpler, and trying to load the file all at once in the constructor seemed to fail for some reason.

I'm raising this draft PR to see if anyone has any ideas about how this could be improved whilst still giving us an easy way to have env files for multiple backend environments. However, if we can think of a better way to approach it then I guess we should leave it as it currently is and scrap this PR.

With this approach, an env.local.json file can be added to the Resources folder, and then GITIGNORED (although I haven't added it to the GITIGNORE yet) and it will use those values instead of the default ones.


Runtime/Env/Env.cs
HarrisonHough, line 1: I don't particularly like the idea of having to force load a file. I find it can make the editor seem slow. 
I would still prefer something simpler, like using Scripting define symbols. 
EG

`
#if !RPM_DEV_ENV
    // prod env urls
    public static class Env
    {

        public static string RPM_API_SUBDOMAIN_URL => "Example.Url";
        public static string RPM_API_V2_URL => "Example.Url";
        public static string RPM_API_V1_URL => "Example.Url";

    }
#endif
    
#if RPM_DEV_ENV
    // dev env urls in a separate class only present in our dev project? 
    public static class Env
    {

        public static string RPM_API_SUBDOMAIN_URL => "Example.Dev.Url";
        public static string RPM_API_V2_URL => "Example.Dev.Url";
        public static string RPM_API_V1_URL => "Example.Dev.Url";

    }
#endif
`

MaxAndreassenRPM: Yup, I agree. Do think we should just go ahead and do this then? 🤔

Runtime/Env/EnvLoader.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace ReadyPlayerMe.AvatarCreator
+{
+    public static class EnvLoader
+    {
+        private const string LOCAL_ENV_FILENAME = "env.local";
+        
+        public static string LoadVar(string varName, string defaultValue)
+        {
+            var jsonContent = Resources.Load<TextAsset>(LOCAL_ENV_FILENAME);
```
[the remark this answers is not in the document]
MaxAndreassenRPM: The SDK wouldn't break, because this file isn't actually required and wouldn't be included for any developers (it's purely a local thing that would be GITIGNORED). Still though, your other comment makes more sense to me, even if it does mean we have to have the values hardcoded for partners too. The original goal was that partners would never see any dev env vars.

Runtime/Env/Env.cs
rYuuk: I agree, though this will allow other developers to use are dev endpoint as well.

MaxAndreassenRPM: Yeah, I'm also apprehensive about that too @rYuuk, but maybe the pros outweigh the cons? 🤔

rYuuk: I guess the dev endpoint is already public, so it should be fine and we can also limit request to it based on some authentication which is only allowed on develop(though I don't know how that will work 😅)

MaxAndreassenRPM: Yeah, maybe let's just keep it simple and go for this then and see what happens? 🔥

Runtime/Env/EnvLoader.cs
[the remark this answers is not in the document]
HarrisonHough: RIght I think I misunderstood. So this EnvLoader.cs would only be in our Unity-SDK project not in Core? 🤔

MaxAndreassenRPM:
Closing this PR as I think we've decided to go with a simpler solution and leaving it open here is causing confusion - a new PR will one day come with the simpler solution 👍

PR #173, opened 2023-11-30
Feature: Uniform request naming

MaxAndreassenRPM:
Base on our discussion yesterday, I've updated the naming of the base requests to be more uniform and reflective of the actual API resources.

image

### DIFF
--- a/Runtime/AvatarCreator/WebRequests/AuthorizedRequest.cs
+++ b/Runtime/AvatarCreator/WebRequests/AuthorizedRequest.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading;
+﻿using System;
+using System.Threading;
 using System.Threading.Tasks;
 using ReadyPlayerMe.Core;
 using UnityEngine.Networking;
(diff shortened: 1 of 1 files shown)


Runtime/AvatarCreator/WebRequests/AuthorizedRequest.cs
```diff
@@ -23,15 +21,8 @@ public async Task<T> SendRequest<T>(RequestData requestData, CancellationToken c
...
-                {
-                    await AuthManager.RefreshToken();
-                }
-                catch (Exception)
-                {
-                    throw;
-                }
-                
+                await AuthManager.RefreshToken();
+
```
rYuuk, line 25: I would suggest not removing this, if I remember correctly that this was added for when even RefreshToken request fails and then without it the error is not caught and propagated above. But somehow rider always suggest to remove it.

rYuuk (approved):
🟢 Thanks this makes api more consistent. Please also have a look at a small concern.

Runtime/AvatarCreator/WebRequests/AuthorizedRequest.cs
MaxAndreassenRPM: Oh, interesting, it seemed to be redundant but if it's a weird edge case then I'll add it back 👍

MaxAndreassenRPM: Added back, and also added a comment on it too so others don't make the same assumption in the future that I did.

PR #174, opened 2023-11-30
Feature/endpoint class refactor

MaxAndreassenRPM:
Here is a draft PR for removing the concept of endpoint files whose sole purpose is to define the URLs for requests. In this PR, each lowest level SDK method fully encapsulates what it needs to make a request to the API, and so you only have to look in 1 place to understand how a request works. Furthermore, adding future requests now only requires you to add 1 new function, rather than update 2 classes each time.

This is PR is only a draft, as there was still some debate about this. However, I wanted to create a PR to illustrate how it would work. There is still other work to do, but my hope is that this is another step towards make the project structure of the SDK easier to digest and understand.

image


Runtime/AvatarCreator/WebRequests/AvatarAPIRequests.cs
```diff
@@ -118,7 +118,7 @@ public async Task<AvatarProperties> GetAvatarMetadata(string avatarId)
             var response = await authorizedRequest.SendRequest<Response>(
                 new RequestData
                 {
-                    Url = AvatarEndpoints.GetAvatarMetadataEndpoint(avatarId),
+                    Url = $"{Env.RPM_API_V2_BASE_URL}avatars/{avatarId}.json",
```
HarrisonHough, line 121: now I actually see it in action I don't particularly love it. I think it becomes harder to maintain and more error prone. 
If you look at this particular url part  
`{Env.RPM_API_V2_BASE_URL}avatars/`
it's kinda copy pasta everywhere and also done so inconsistently. Some have the "/" before the avatars, others have "/" after it. 
image

One thing that may at least help here is making a constant 
EG
```cs
const string V2_Avatars = $"{Env.RPM_API_V2_BASE_URL}avatars/"
```

MaxAndreassenRPM: Yeah, I only did it quickly for the draft PR purposes, but would need to check and test the links before making it not a draft PR 🙏

I wouldn't be against having a constant at the top of the class for the base url which pertains to the class that it's in. What makes you think it's harder to maintain?

Endpoint URLs pretty much never change, so for something so static I don't understand what the maintenance issue would be 🤔

Runtime/AvatarCreator/WebRequests/AvatarAPIRequests.cs
```diff
@@ -209,10 +222,15 @@ public async Task<string> SaveAvatar(string avatarId)
 
         public async Task DeleteAvatar(string avatarId, bool isDraft = false)
```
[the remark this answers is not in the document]
MaxAndreassenRPM: Yup, I thought the same thing about adding another method (since they are technically different endpoints) but I didn't want to change any specific logic in this PR outside of removing the endpoint files, since that is what I wanted to demonstrate.

Runtime/AvatarCreator/Utils/AvatarRenderHelper.cs
```diff
@@ -11,7 +11,7 @@ public static class AvatarRenderHelper
         public static async Task<Texture2D> GetPortrait(string avatarId, CancellationToken token = default)
         {
             var webRequestDispatcher = new WebRequestDispatcher();
-            var response = await webRequestDispatcher.SendRequest<ResponseTexture>(AvatarEndpoints.GetRenderEndpoint(avatarId), HttpMethod.GET,
+            var response = await webRequestDispatcher.SendRequest<ResponseTexture>($"{Env.RPM_MODELS_BASE_URL}/{avatarId}.png", HttpMethod.GET,
```
MaxAndreassenRPM, line 14: A small improvement I'd now make (but haven't done here yet) is to move this into the main avatar api class, since this class isn't a helper, it's just a class with another avatar api request in it.

Runtime/AvatarCreator/WebRequests/AvatarAPIRequests.cs
MaxAndreassenRPM: Yup, constants at the tops of the files make sense - added 👍

HarrisonHough (approved):
Looks good. Cool stuff!

Runtime/AvatarCreator/WebRequests/AuthAPIRequests.cs
```diff
@@ -9,20 +9,21 @@ public class AuthAPIRequests
     {
         private readonly string domain;
         private readonly IDictionary<string, string> headers = CommonHeaders.GetHeadersWithAppId();
+        private readonly string rpmAuthBaseUrl;
 
         private readonly WebRequestDispatcher webRequestDispatcher;
 
         public AuthAPIRequests(string domain)
```
HarrisonHough, line 16: Is this domain = subdomain? and if so why is it not using core settings subdomain that is used everywhere else?

HarrisonHough (approved):
Overall it looks fine and seems to work, one question about the API request domain param in constructor. Not sure if I am misinterpreting the use of the domain.

Runtime/AvatarCreator/WebRequests/AuthAPIRequests.cs
MaxAndreassenRPM: It seems to be initialized with that value (see screenshot below):

I can update it if you think it makes sense to.

image

PR #175, opened 2023-11-30
Feature/asset panel element

HarrisonHough:
## Description

-   Still in draft but making PR to get feedback, it's quite difficult to do it in a super clean way while still allowing it to be as flexible as possible so I reluctantly went with using inheritance and an interface. I also reworked some things like how colors were handled.
But anyways take a look and share your thoughts. I'm still quite open to ideas and suggestions. 

## How to Test

## Checklist

-   [ ] Changelog is updated.


Runtime/AvatarCreator/Data/AvatarTemplateData.cs
```diff
@@ -6,12 +6,13 @@
 namespace ReadyPlayerMe.AvatarCreator
 {
     [Serializable]
-    public class AvatarTemplateData
+    public class AvatarTemplateData : IAssetData
     {
+        public string Id { get; set; }
+        public Category Category { get; set; }
```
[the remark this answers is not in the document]
MaxAndreassenRPM: Maybe if you changed this to be something like an 'AssetType' enum, then it would make more sense, and then you could hard code this to always be of type 'AvatarTemplate' with something like AssetType AssetType => AssetType.AvatarTemplate?

Runtime/AvatarCreator/WebRequests/PartnerAssetsRequests.cs
```diff
@@ -69,6 +69,48 @@ public async Task<PartnerAsset[]> Get(BodyType bodyType, OutfitGender gender, Ca
...
         }
 
+        public async Task<PartnerAsset[]> GetCategory(Category category, BodyType bodyType, OutfitGender gender, CancellationToken ctx = new CancellationToken())
+        {
+            var assets = new HashSet<PartnerAsset>();
+            AssetLibrary assetLibrary;
+
+            try
+            {
+                assetLibrary = await GetRequest(LIMIT, 1, category, gender, bodyType, ctx: ctx);
```
MaxAndreassenRPM, line 79: I know it's not in your PR, but can we rename 'GetRequest' to something more explicit, as it currently means nothing 😅 - maybe 'ListAssetsRequest'?

Runtime/AvatarCreator/Data/AvatarTemplateData.cs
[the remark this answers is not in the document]
HarrisonHough: Yeah i'm not proud of that 😅
Perhaps AssetType atually does make more sense 🤔 
or alternatively insteead of using an interface and forcing a "categorization" I could have some sort of CustomizationData class (similar to AssetData) and just make a converter to convert from the different types. (TemplateAvatar, PartnerAsset, Colors etc)

Runtime/AvatarCreator/WebRequests/PartnerAssetsRequests.cs
```diff
@@ -69,7 +69,49 @@ public async Task<PartnerAsset[]> Get(BodyType bodyType, OutfitGender gender, Ca
             return assets.ToArray();
         }
 
-        public async Task<PartnerAsset[]> Get(Category? category, BodyType bodyType, OutfitGender gender, CancellationToken ctx = new CancellationToken())
+        public async Task<PartnerAsset[]> GetCategory(AssetType assetType, BodyType bodyType, OutfitGender gender, CancellationToken ctx = new CancellationToken())
+        {
+            var assets = new HashSet<PartnerAsset>();
```
rYuuk, line 74: Maybe both the function(Get and GetCategory) in this class can be merge into one?

Runtime/AvatarCreator/UI/Elements/SelectionElement.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    public abstract class SelectionElement : MonoBehaviour
+    {
+        [Header("UI Elements")]
+        [Space(5)]
+        [SerializeField] private ButtonElement buttonElementPrefab;
+        [SerializeField] private Transform buttonContainer;
+        [SerializeField] private GameObject selectedIcon;
+        public UnityEvent<IAssetData> onAssetSelected;
+        private Dictionary<string, ButtonElement> buttonMap = new Dictionary<string, ButtonElement>();
```
rYuuk, line 21: I was using map initially, but I don't like it. Nowadays I prefer something like buttonElementById which sound much more descriptive.

Runtime/AvatarCreator/WebRequests/PartnerAssetsRequests.cs
HarrisonHough: yeah I actually don't remember why I did this 😅

Runtime/AvatarCreator/IconFetcher.cs
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public async Task<Texture> GetIcon(string url)
+    {
+        url = $"{url}?w={size.ToString()}";
+        Debug.Log($"Get icon {url}");
+        var downloadHandler = new DownloadHandlerTexture();
+        var response = await webRequestDispatcher.SendRequest<ResponseTexture>(url, HttpMethod.GET, downloadHandler: downloadHandler, ctx: ctx);
+        response.ThrowIfError();
+        return response.Texture;
```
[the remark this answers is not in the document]
HarrisonHough: thanks! have updated this now and removed iconfetcher

Runtime/AvatarCreator/UI/Elements/SelectionElement.cs
```diff
@@ -0,0 +1,69 @@
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Events;
+
+namespace ReadyPlayerMe.AvatarCreator
+{
+    public interface IAssetData
+    {
+        public string Id { get; set; }
+        public AssetType AssetType { get; set; }
```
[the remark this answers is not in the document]
HarrisonHough: yep have done so now. Although probably I should move it into a better subdirectory 🤔

Runtime/AvatarCreator/PartnerAssetFetcher.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace ReadyPlayerMe.AvatarCreator
+{
+    public class PartnerAssetFetcher
+    {
+        private readonly PartnerAssetsRequests partnerAssetsRequests;
+
+        public async Task<PartnerAsset[]> FetchAssetData(Category category, BodyType bodyType, OutfitGender gender, CancellationToken token = default)
+        {
+            return await partnerAssetsRequests.Get(category, bodyType, gender, token);
```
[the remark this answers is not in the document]
HarrisonHough: right so this actually wasn't used :D I started making it but then never used and forgot to delete.

Runtime/AvatarCreator/Data/ColorPalette.cs
```diff
@@ -2,16 +2,28 @@
 
 namespace ReadyPlayerMe.AvatarCreator
 {
+
+    [Serializable]
+    public struct ColorLibrary
```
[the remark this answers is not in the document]
HarrisonHough: done

Runtime/AvatarCreator/UI/Elements/SelectionElement.cs
HarrisonHough: updated

Runtime/AvatarCreator/Data/ColorLibrary.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+
+namespace ReadyPlayerMe.AvatarCreator
+{
+
```
MaxAndreassenRPM, line 5: nitpick: remove random space

rYuuk (approved):
🟢 Great Job in figuring this big one out 💯 🥇 🎊

Runtime/AvatarCreator/UI/Elements/AssetSelectionElement.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        /// <summary>
+        /// Asynchronously loads partner asset data based on the specified asset type, body type, and gender.
+        /// This method updates the internal partnerAssets collection with the fetched data.
+        /// </summary>
+        /// <param name="bodyType">The body type to filter the assets and determine the type of avatar to load.</param>
+        /// <param name="gender">The gender to filter the assets and determine which skeleton will be loaded.</param>
+        /// <returns>A Task representing the asynchronous operation of fetching and loading the partner asset data.</returns>
+        public async Task LoadTemplateData(OutfitGender gender)
```
MaxAndreassenRPM, line 41: This method's name should be changed, as it isn't template data in this context.

MaxAndreassenRPM (changes requested):

Runtime/AvatarCreator/UI/Elements/AssetSelectionElement.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// It extends the functionality of SelectionElement to handle PartnerAsset-specific interactions.
+    /// </summary>
+    public class AssetSelectionElement : SelectionElement
+    {
+        [Header("Properties")]
+        [SerializeField] private BodyType bodyType = BodyType.FullBody;
+        [SerializeField] private AssetType assetType;
+        [SerializeField] private int iconSize = 64;
+
+        private PartnerAsset[] partnerAssets;
```
MaxAndreassenRPM, line 21: Could we just call this 'asset'? PartnerAsset implies it's some sort of join object between a Partner (or application) and an Asset. Actually this object has no property that references the partner so the naming doesn't make much sense IMO.

Runtime/AvatarCreator/UI/Elements/AssetSelectionElement.cs
HarrisonHough: yep! good catch

HarrisonHough: updated

Runtime/AvatarCreator/UI/Elements/AssetSelectionElement.cs
HarrisonHough: updated

MaxAndreassenRPM:
Tested locally with the TEMP scene and it works well 👍

A couple of small things I'd suggest we could improve on:

1. Adding a default loader to API requests
2. Working out how to make it obvious that the asset list element doesn't work out of the box (maybe an error mentioning what is required, or an example scene in the Samples folder, or even some slight re-work so it can at least download assets out of the box).

Either way, it is good as it is, but the above can be improved in the future.

MaxAndreassenRPM (approved):

PR #176, opened 2023-11-30
chore: reduce repeated code

MaxAndreassenRPM:
An idea to reduce the duplication between the bespoke asset panel types.

### DIFF
--- a/Runtime/AvatarCreator/UI/Elements/SelectionElement.cs
+++ b/Runtime/AvatarCreator/UI/Elements/SelectionElement.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Collections.Generic;
+using ReadyPlayerMe.Core;
 using UnityEngine;
 using UnityEngine.Events;
 
@@ -12,6 +14,8 @@ namespace ReadyPlayerMe.AvatarCreator
     
     public abstract class SelectionElement : MonoBehaviour
     {
+        private const string TAG = nameof(SelectionElement);
+        
         [Header("UI Elements")]
         [Space(5)]
         [SerializeField] private ButtonElement buttonElementPrefab;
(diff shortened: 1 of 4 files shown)


Runtime/AvatarCreator/UI/Elements/TemplateSelectionElement.cs
```diff
@@ -26,55 +27,27 @@ private void Awake()
         public async void LoadAndCreateButtons()
         {
             await LoadTemplateData();
-            CreateButtons();
-            await LoadTemplateRenders();
+            CreateButtons(avatarTemplates.ToArray(), async (button, asset) =>
+            {
+                var downloadHandler = new DownloadHandlerTexture();
```
MaxAndreassenRPM, line 32: We really need to wrap this into some sort of reusable image download request class. I know it was mentioned we had one somewhere but I couldn't find it 😅

Runtime/AvatarCreator/UI/Elements/SelectionElement.cs
```diff
@@ -29,6 +33,23 @@ public ButtonElement CreateButton(string id)
             return button;
         }
 
+        public void CreateButtons<T>(T[] assets, Action<ButtonElement, T> onButtonCreated = default) where T : IAssetData
```
HarrisonHough, line 36: don't you think that this make it look more complicated (and potentially harder to write simple documentation for) with the use of generics and function passing? 
I understand that you wanted to remove some code duplication though.

rYuuk: I like generics in general, but this could be overengineering, though I don't know any other way to remove duplication right now.

MaxAndreassenRPM: Personally, I don't really see it as an issue, because it is a generic in a lower level part of the element? The core documentation for partners would be that they could easily extend this class with their own functions to create their own selector, and they get a helpful callback to edit the button style if they want to.

IMO this is easier to understand and extend than it would be the other way around since there's less uniformity otherwise.

In my mind this strips away the code that the partners don't **need** care about, but of course they can if they want to.

PR #177, opened 2023-12-01
refactor: package event listener

rYuuk:
## Description

- The model namespace caused PackageCoreInfo.cs to no come in same asmdef, hence causing error when it was used outside of this asmdef.
- I made it simpler by removing redundant code and using simple string type.


Editor/Utils/PackageManager/PackageManagerEventListener.cs
```diff
@@ -1,28 +1,32 @@
...
 
         static void OnPackagesInstalled(PackageRegistrationEventArgs packageRegistrationEventArgs)
         {
             packageRegistrationEventArgs.added
-                .Select(package => new PackageCoreInfo
-                {
-                    Id = package.packageId,
-                    Name = package.displayName,
-                })
+                .Where(packageInfo => packageInfo.packageId.Contains(REQUIRED_KEYWORD))
```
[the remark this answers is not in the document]
rYuuk: but then it will be fired for any package installed doesn't matter if its our partners or random 3rd party package which could cause a clutter in analytics events.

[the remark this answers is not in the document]
rYuuk: Though if that's the intention I can remove this check.

Editor/Analytics/AmplitudeEditorLogger.cs
```diff
@@ -234,13 +233,12 @@ public void LogAvatarCreatorSampleImported()
             LogEvent(EventName.AVATAR_CREATOR_SAMPLE_IMPORTED);
         }
 
-        public void LogPackageInstalled(PackageCoreInfo packageInfo)
+        public void LogPackageInstalled(string id, string name)
```
MaxAndreassenRPM, line 236: You're not a fan of DTOs are you? 😅 -> https://www.c-sharpcorner.com/article/data-transfer-objects-dtos-in-c-sharp/

Fair enough though, as long as we agree on a consistent approach then I'm happy with whatever convention we use. Do we have a limit on the amount of properties that we pass to a function before we would wrap it in an object?

rYuuk: I am which as you see there are way too many structs and class object holding data, but here I didn't wanted Analytics to have a cross dependency on PackageCoreInfo.

rYuuk: Also as mentioned before the model namespace caused an issue.

Editor/Utils/PackageManager/PackageManagerEventListener.cs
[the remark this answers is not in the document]
HarrisonHough: Do we need to worry about user permission in regards to tracking every package a developer installs? 😅

Editor/Analytics/AmplitudeEditorLogger.cs
HarrisonHough: I believe this was don't mostly because of issues related to using the PackageCoreInfo object right?
We also weren't using all the props anyways but thats kinda minor

Editor/Utils/PackageManager/PackageManagerEventListener.cs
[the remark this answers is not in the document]
MaxAndreassenRPM: It's not sensitive data, since we're just checking the package name, so I don't see why we would have a user permission problem - it's only if they opt into analytics which I believe is enough 🤔

> but then it will be fired for any package installed doesn't matter if its our partners or random 3rd party package which could cause a clutter in analytics events.

This could be super valuable though, maybe there's a package that's often used with RPM which we have no idea about. If so then we could reach out to them and maybe add them to the integrations page.

Editor/Analytics/AmplitudeEditorLogger.cs
MaxAndreassenRPM: Yup, agreed that everything at the moment is using primitives so makes sense to keep it uniform, like you've done 🙂

Editor/Analytics/AmplitudeEditorLogger.cs
```diff
@@ -234,13 +233,12 @@ public void LogAvatarCreatorSampleImported()
...
         {
             LogEvent(EventName.INSTALL_PACKAGE, new Dictionary<string, object>
             {
-                { "id", packageInfo.Id },
-                { "name", packageInfo.Name },
-                { "url", packageInfo.Url },
```
MaxAndreassenRPM, line 243: Nice spot to drop the url value which is never populated now. I was going to question it at first, but then remembered we now get the RPM package url version in the user properties ✅

Editor/Utils/PackageManager/PackageManagerEventListener.cs
```diff
@@ -1,28 +1,35 @@
...
-    public class PackageManagerEventListener
+    public abstract class PackageManagerEventListener
     {
+        public static event Action<string> OnPackageImported;
+
         [InitializeOnLoadMethod]
         static void Initialize()
         {
-            Events.registeringPackages += OnPackagesInstalled;
+            Events.registeredPackages += OnPackagesInstalled;
```
MaxAndreassenRPM, line 16: Are you sure this change was needed? 🤔 

https://docs.unity3d.com/2023.3/Documentation/ScriptReference/PackageManager.Events-registeringPackages.html

HarrisonHough (approved):
Code looks fine and seems to work. I mentioned this to you already but I thought that it also worked before with the initial implementation, at least when I remember tested 😅. 
It was hard to find the events in Amplitude seems there is some delay with the sync or something

MaxAndreassenRPM:
> Code looks fine and seems to work. I mentioned this to you already but I thought that it also worked before with the initial implementation, at least when I remember tested 😅. I remember complaining to Max about firing multiple events but then I realized it was because of the multiple packages or whatever

It was hard to find the events in Amplitude seems there is some delay with the sync or something

Yup, I remember the same - I spoke to @rYuuk about it and we were trying to understand if there was something about his local setup that prevented the other event from working for him.

HarrisonHough:
> > Code looks fine and seems to work. I mentioned this to you already but I thought that it also worked before with the initial implementation, at least when I remember tested 😅. I remember complaining to Max about firing multiple events but then I realized it was because of the multiple packages or whatever
> 
> Yup, I remember the same - I spoke to @rYuuk about it and we were trying to understand if there was something about his local setup that prevented the other event from working for him, because I also went back to look at past Amplitude events and it definitely was also working previously too.

not sure if relevant here but when calling stuff in editor constructors I find that behaviour for order of operations can differ per machine. In particular I notice the order is different on a mac. But there are also some errors or logs that show in different order depending on whether I use my personal laptop or work one 😅

PR #178, opened 2023-12-04
[SDK-681] Add account creation element

rYuuk:
## Description

- Added account creation element scripts and prefab
- Fixed wrong url in login with otp

## How to Test

-  NA

## Checklist

-   [ ] Changelog is updated.

[SDK-681]: https://ready-player-me.atlassian.net/browse/SDK-681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HarrisonHough:
A minor thing but the prefabs buttons has a bunch of empty events assigned. 
image

Runtime/AvatarCreator/UI/Elements/AccountCreationElement.cs
```diff
@@ -24,27 +23,24 @@ private void OnEnable()
         {
             sendEmailButton.onClick.AddListener(OnSendEmailButton);
```
HarrisonHough, line 24: In theory we could make this slightly more flexible by removing the auto assigning of the event listeners and instead make the appropriate functions public and assign them in the inspector. It would also mean that you might need the button references. But not a blocker for me if you think it's unnecessary to do that.

HarrisonHough (approved):
Looks good! some small things in the comments

PR #181, opened 2023-12-05
[SDK-685] Feature/avatar preview element

HarrisonHough:
## Description

-   Adds an avatar preview component which combines the avatar rotation script and a new CameraFocuser script (open to better name suggestions for camera script). 
- I wasn't planning to change the way the Avatar rotation and input handling work but I think this new approach would be good as it enables devs to use other input sources (EG keyboard or controller) by using the IAvatarRotatorInput interface. But please give feedback on this 
- I also decided to make the avatar something that could be set, so that it can be added on a component other than on an avatar and still work. But again open to other opinions here 

Component looks like this 
image

Made a temp demo scene to make it easy to trigger the camera focus stuff  if you check our the avatar-preview feature branch on Unity-SDK repo you can test for yourself
image

## How to Test

-   if you check our the avatar-preview feature branch on Unity-SDK repo you can test for yourself in scene located in TEMP folder

## Checklist

-   [ ] Changelog is updated.


Samples~/AvatarCreatorSamples/Scripts/UI/SelectionScreens/AvatarCreatorSelection.cs
```diff
@@ -301,8 +301,9 @@ private void ProcessAvatar(GameObject avatar)
                 avatar.GetComponent<Animator>().runtimeAnimatorController = animator;
             }
             avatar.transform.rotation = lastRotation;
-            avatar.AddComponent<MouseInput>();
-            avatar.AddComponent<AvatarMouseRotation>();
+            avatar.AddComponent<MouseRotationHandler>();
+            var rotator = avatar.AddComponent<AvatarRotator>();
+            rotator.SetAvatar(avatar);
```
rYuuk, line 306: doesn't this feels redundant if component is attached to avatar, it will have access to avatar then why rotatar needs to be set.

rYuuk (approved):
🟢 Tested, it looks good. A small comment.

PR #199, opened 2024-01-03
Feature/remove avatar shadow cac

rYuuk:
## Description

## How to Test

## Checklist

-   [ ] Changelog is updated.


HarrisonHough (approved):
Overall the code looks good. 
1 improvement idea is that we could look at showing only shader properties (or rather texture channels) that have been enabled? 
image
EG in this case only BaseColor would be visible under the Shader section

Also as mentioned in slack but maybe it makes sense to explore the gltfast instantiator approach for assigning materials/shaders to prevent the "double" assigning. Or at least test if doing this has any benefit performance wise

PR #206, opened 2024-01-15
feat: don't save, when avatar is not in draft

rk132:
## Description

-  When there is no draft avatar, then don't save the avatar.

## How to Test

-   Try to go over the old avatar creator with an update on the avatar (change some assets) and without changing and verify, that the avatar is successfully loaded in both cases.

[SDK-401]: https://ready-player-me.atlassian.net/browse/SDK-401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HarrisonHough (approved):
Looks good and seems to work. 
Great stuff!

PR #208, opened 2024-01-18
[SDK-744] - Remove webview from the core

rk132:
## Description

-   Remove the Webview package installation from the core

## How to Test

-   Import the core package and observe, that the Webview is not installed anymore.

[SDK-744]: https://ready-player-me.atlassian.net/browse/SDK-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Editor/Core/Scripts/ModuleManagement/ModuleList.cs
+++ b/Editor/Core/Scripts/ModuleManagement/ModuleList.cs
@@ -18,6 +18,20 @@ namespace ReadyPlayerMe.Core.Editor
         };
 
         /// <summary>
+        ///     A static list of all the required modules represented in an array of <c>ModuleInfo</c>.
+        /// </summary>
+        public static readonly ModuleInfo[] Modules =
+        {
+            new ModuleInfo
+            {
+                name = "com.readyplayerme.webview",
+                gitUrl = "https://github.com/readyplayerme/rpm-unity-sdk-webview.git",
+                branch = "",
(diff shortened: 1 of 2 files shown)


Editor/Core/Scripts/ModuleManagement/ModuleList.cs
```diff
@@ -53,14 +39,6 @@ public static class ModuleList
             var installedModules = new Dictionary<string, string>();
             installedModules.Add(Core.name, Core.version);
 
-            foreach (ModuleInfo module in Modules)
```
HarrisonHough, line 56: I think removing this (and WebView) from module list will cause us to lose tracking of whether WebView has been installed.

HarrisonHough (changes requested):
Overrall it looks fine however you might need to add back the WebView to module list so that we can track whether or not it is installed? 
Unless we try do it some other way from inside WebView package itself

Editor/Core/Scripts/ModuleManagement/ModuleList.cs
rk132: Yeah, makes sense to have it as it is already in place

HarrisonHough (approved):
Looks good now, nice work.

PR #211, opened 2024-01-31
[SDK-780] Feature/simple cac updates

HarrisonHough:
## Description

-   added logic to reset/fix the issue that icons loaded/set while the object was disabled would not correctly resize to fit the button.

image

## How to Test

-   easiest way is to run the simple creator example, as soon as it starts switch to headwear (before outfits have time to load). 
Then wait a few seconds and switch back to outfit panel. If the icons are sized correctly then the fix is working. 

## Checklist

-   [ ] Changelog is updated.

[SDK-780]: https://ready-player-me.atlassian.net/browse/SDK-780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Runtime/AvatarCreator/Scripts/UI/Buttons/SelectionButton.cs
```diff
@@ -27,10 +29,16 @@ public void AddListener(Action action)
         /// </summary>
         /// <param name="texture">The texture to be assigned to the RawImage component</param>
         /// <param name="sizeToParent">If true the icon will resize itself to fit inside the parent RectTransform</param>
-        public void SetIcon(Texture texture, bool sizeToParent = true)
+        public void SetIcon(Texture texture)
         {
+            if (rawImageRectTransform == null)
+            {
+                rawImageRectTransform = rawImage.GetComponent<RectTransform>();
+            }
```
rk132, line 37: What about converting variable to property and defining this inside of the property:

```
private Rigidbody m_rigidbody = null;
 
private Rigidbody rigidbody {
   get {
     if (!m_rigidbody)
       m_rigidbody = GetComponent<Rigidbody>();
 
     return m_rigidbody;
   }
}
```

Samples~/AvatarCreatorSamples/AvatarCreatorWizard/Scripts/UI/Buttons/AvatarButton.cs
```diff
@@ -16,13 +16,15 @@ public class AvatarButton : MonoBehaviour, IPointerEnterHandler, IPointerExitHan
...
 
+        private RectTransform rawImageRectTransform;
         private string avatarId;
         private bool showButtons;
 
         private CancellationTokenSource ctxSource;
 
         private async void Start()
         {
+            rawImageRectTransform = image.GetComponent<RectTransform>();
```
rk132, line 27: What about converting variable to property and defining this inside of the property.

Samples~/AvatarCreatorSamples/AvatarCreatorWizard/Scripts/UI/Buttons/AssetButton.cs
```diff
@@ -25,8 +27,14 @@ public void SetColor(string colorHex)
 
         public void SetIcon(Texture texture)
         {
+            if (rawImageRectTransform == null)
+            {
+                rawImageRectTransform = icon.GetComponent<RectTransform>();
```
rk132, line 32: What about converting variable to property and defining this inside of the property:

rk132 (approved):
Looks good, but let me know wdyt about the questions above.

Runtime/AvatarCreator/Scripts/UI/Buttons/SelectionButton.cs
HarrisonHough: It's not a bad idea, but I do think it is kinda overkill and makes the code a bit less readable. 
If we had other functions that made use of the RectTransform it would make more sense, but at the moment it is isolated only to the SetIcon function.

rk132: got it, i don't have strong feelings about this so feel free to merge.

PR #213, opened 2024-02-01
[SDK-776] Feature/default asset selection

HarrisonHough:
## Description

-   Added default asset selection for template avatars

NOTE this only works for some assets, it doesn't work for outfits which is the same in web 

## How to Test

-   Run the Simple Avatar craetor (Avatar Creator Experimental) and check that after initial load the hair and eyebrows are selected already 

## Checklist

-   [ ] Changelog is updated.

[SDK-776]: https://ready-player-me.atlassian.net/browse/SDK-776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rk132 (approved):
Looks good. You still added the headwear and outfit to the onAvatarCreated list even tho they don't work right now.  Should we remove it from the Scene list then?

PR #214, opened 2024-02-01
feat: add rpm-source header to the requests

rk132:
## Description

-   Add rpm-source header to the requests. This is used to track the usage of avatar creator.

## How to Test

-   Right now save draft endpoint processes this data. Check, that if you save the draft in the avatar creator, that in amplitude, there will be one extra entry. (This is cached for 12 hours, so you can make only one request)

[SDK-627]: https://ready-player-me.atlassian.net/browse/SDK-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HarrisonHough (approved):
Looks good!

PR #218, opened 2024-02-07
Feature/add avatar list element

rk132:
## Description

-   Create element for Avatar Selection and Management

## How to Test

-   Add login and new avatar list element. After that log in and test, that elements are listed for your account, and that they can be deleted.


Runtime/AvatarCreator/Scripts/UI/Elements/AvatarListElement.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            if (!AuthManager.IsSignedIn && !AuthManager.IsSignedInAnonymously)
+            {
+                SDKLogger.Log(TAG, "Not signed in");
+                return;
+            }
+
+            avatarAPIRequests ??= new AvatarAPIRequests();
+
+            var avatarPartnerArr = await avatarAPIRequests.GetUserAvatars(AuthManager.UserSession.Id);
```
HarrisonHough, line 45: Minor naming suggestion, since it a dictionary I suggest maybe following the convention of valueByKey for naming. 
EG partnerByAvatarId or something like that

```suggestion
            var partnerByAvatarId = await avatarAPIRequests.GetUserAvatars(AuthManager.UserSession.Id);
```
Kinda similar to what we have in SelectionElement 
image

Runtime/AvatarCreator/Scripts/UI/Elements/AvatarListElement.cs
```diff
@@ -0,0 +1,112 @@
...
+        private enum AvatarListFilter
+        {
+            Application,
+            None
+        }
+
+        private const string TAG = nameof(AvatarListElement);
+
+        [SerializeField] private Transform avatarsContainer;
+        [FormerlySerializedAs("avatarListButton")]
```
HarrisonHough, line 24: this can be removed

Runtime/AvatarCreator/Scripts/UI/Elements/AvatarListElement.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using ReadyPlayerMe.Core;
+using UnityEngine;
+using UnityEngine.Events;
+using UnityEngine.Serialization;
+using UnityEngine.UIElements;
```
HarrisonHough, line 8: Keep an eye out on the unused using statements in Rider (greyed out lines) as they can all be removed 
image

HarrisonHough (changes requested):
Some minor suggestions and things to update. 
Also I think you should reorder these function calls so that you run the SetupDeletion After it is Set Active, the reason being that functions don't always run when they are called on an object/class that is disabled.
image

rk132:
> Some minor suggestions and things to update. Also I think you should reorder these function calls so that you run the SetupDeletion After it is Set Active, the reason being that functions don't always run when they are called on an object/class that is disabled. image

I think, only the events, that are related to the main thread (Awake, Update, FixedUpdate, Start, etc...) are not called when the object is not active. When I call a function directly against the monobehaviour, then it shouldn't matter as far as I know.

Runtime/AvatarCreator/Scripts/UI/Buttons/AvatarListItem.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            deleteAvatarButton.onClick.AddListener(deleteAvatar.Invoke);
+        }
+    }
+
+    /// <summary>
+    /// Sets the icon of the avatar component
+    /// </summary>
+    /// <param name="texture">The texture to be assigned to the RawImage component</param>
+    /// <param name="sizeToParent">If true the icon will resize itself to fit inside the parent RectTransform</param>
```
HarrisonHough, line 43: wrong param name. Think you forgot you update the param name. Also I think this function should be called something more like LoadIcon.

Runtime/AvatarCreator/Scripts/UI/Buttons/AvatarListItem.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        if (avatarImage == null)
+        {
+            return;
+        }
+        var texture = await AvatarRenderHelper.GetPortrait(avatarId);
+
+        var avatarRectTransform = avatarImage.GetComponent<RectTransform>();
+
+        var previousSize = avatarRectTransform.sizeDelta;
```
HarrisonHough, line 54: maybe instead of this you can use the aspectRatio fitter like I did for the selection button 
https://github.com/readyplayerme/rpm-unity-sdk-core/blob/7a5cb4bb1f2e4c380c3b92d8fdbbb6ac24078f10/Runtime/AvatarCreator/Scripts/UI/Buttons/SelectionButton.cs#L37

Runtime/AvatarCreator/Scripts/UI/Buttons/AvatarListItem.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+using ReadyPlayerMe.AvatarCreator;
+using UnityEngine;
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+public class AvatarListItem : MonoBehaviour
```
HarrisonHough, line 7: I think a better name for this class would be AvatarButton or AvatarSelectButton.

Runtime/AvatarCreator/Scripts/UI/Buttons/AvatarListItem.cs
```diff
@@ -0,0 +1,59 @@
...
+    public RawImage avatarImage;
+    public UnityEvent onImageLoaded;
+
+    /// <summary>
+    /// Adds the listeners to the button's onClick event.
+    /// </summary>
+    /// <param name="customizeAvatar">A function to run when the customize button is clicked</param>
+    /// <param name="selectAvatar">A function to run when the select button is clicked</param>
+    /// <param name="deleteAvatar">A function to run when the delete button is clicked</param>
+    public void AddListener(Action customizeAvatar, Action selectAvatar, Action deleteAvatar)
```
HarrisonHough, line 21: I actually have an idea for reducing some code duplication here. 
You can add some small data type (like a struct) to hold avatarID and an enum (eg actionType) which would have 3 states Customize, Select, Delete etc then you would only need to pass 1 function AND it becomes more extensible later if there would be some other type of action (not that this is likely). 
I also think that this could then potentially be merged with the LoadIcon function so you only need to calll 1 "setup" function. What do you think?

HarrisonHough (changes requested):
Sorry to be a pain!
Minor things in comments and some refactor/renaming suggestions

Runtime/AvatarCreator/Scripts/UI/Buttons/AvatarListItem.cs
rk132: I can test something

Runtime/AvatarCreator/Scripts/UI/Elements/AvatarListElement.cs
```diff
@@ -0,0 +1,112 @@
...
+                }
+            }
+            partnerByAvatarId.Clear();
+        }
+
+        private void CreateButton(string avatarId)
+        {
+            var avatarListElement = Instantiate(userAvatarElement, avatarsContainer).GetComponent<UserAvatarElement>();
+            avatarListElement.SetupButton(avatarId,
+                new UserAvatarElement.AvatarListItemAction { actionToPerform = () => OnAvatarModified(avatarId), actionType = UserAvatarElement.ButtonAction.Customize },
```
HarrisonHough, line 94: I still think that doing things this way with, passing 3 separate actions is a bit overcomplicated and harder to read. All 3 buttons have the same event arguments (avatarId) the only difference is that the button action.

HarrisonHough (approved):
Good to go. Great work!

PR #223, opened 2024-02-12
feat: update avatar creator simple scene to contain user management, create, and save avatar functionality

rk132:
## [SDK-683](https://ready-player-me.atlassian.net/browse/SDK-683)

image

## Description

-   Update the avatar creator scene, so that you can log in/log out, and create and save new avatars from the template.

## How to Test

-   Verify, that everything works

[SDK-683]: https://ready-player-me.atlassian.net/browse/SDK-683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Samples~/AvatarCreatorSamples/AvatarCreatorExperimental/Scripts/SimpleAvatarCreator.cs
+++ b/Samples~/AvatarCreatorSamples/AvatarCreatorExperimental/Scripts/SimpleAvatarCreator.cs
@@ -23,7 +23,7 @@ namespace ReadyPlayerMe.Samples.AvatarCreatorExperimental
         [SerializeField] private GameObject loading;
 
         [SerializeField] private BodyType bodyType = BodyType.FullBody;
-        private readonly OutfitGender gender = OutfitGender.Masculine;
+        private OutfitGender gender = OutfitGender.Masculine;
 
         private AvatarManager avatarManager;
         private GameObject avatar;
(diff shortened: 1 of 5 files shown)


Samples~/AvatarCreatorSamples/AvatarCreatorExperimental/Scripts/SimpleAvatarCreator.cs
```diff
@@ -137,7 +187,9 @@ private void SetupAvatar()
         {
             avatar.AddComponent<MouseRotationHandler>();
             avatar.AddComponent<AvatarRotator>();
-            avatar.GetComponent<Animator>().runtimeAnimatorController = animationController;
+            var animator = avatar.GetComponent<Animator>();
+            animator.runtimeAnimatorController = animationController;
```
HarrisonHough, line 191: Minor thing but probably better to set the runtimeAnimatorController after calling the SetupAnimator function.

HarrisonHough (approved):
Looks good. 
Not specifically related to this PR but I think there are some things we could add or improve on like: 
- On the Avatars (user avatars) panel it could show some text if it's empty like "Sign in to view avatars" or "no avatars found" 
- When you click "Select Avatar" on a user avatar in this example it actually does nothing. Perhaps we should make it load an avatar? 
- maybe a separate task to improve the UI layout a little. We don't need to focus too much making it pretty but maybe we can organize it in a way that makes it easier to understand. Like a panel with all the buttons and grouping under headings or something

rk132:
> Looks good. Not specifically related to this PR but I think there are some things we could add or improve on like:
> 
> * On the Avatars (user avatars) panel it could show some text if it's empty like "Sign in to view avatars" or "no avatars found"
> * When you click "Select Avatar" on a user avatar in this example it actually does nothing. Perhaps we should make it load an avatar?
> * maybe a separate task to improve the UI layout a little. We don't need to focus too much making it pretty but maybe we can organize it in a way that makes it easier to understand. Like a panel with all the buttons and grouping under headings or something

For the first point, I don't want to introduce any new scripts to the simple avatar creator. I want it to be as simple as possible. So if a developer wants a text there, then they can add it and react against the OnAvatarsLoaded event in the avatarListElement.

For the second point, makes sense. I didn't know as well what to do with that button to be fair 😅. I'll remove it from the list prefab for now as we don't manage "game state" right now. So there's no next state, where we can close the creator and show the avatar.

For the third point, will do 👍

PR #228, opened 2024-02-14
feat: ask, if user wants to continue editing their previous avatar, when they stopped at the middle

rk132:
image

## Description

-   Ask, if the user wants to continue editing their avatar if they closed the avatar creator while editing and open it again

## How to Test

- log in
- Start modifying/creating new avatar
- Close scene (Exit play mode)
- Run again and see, that the pop-up that asks to continue editing is shown

[SDK-723]: https://ready-player-me.atlassian.net/browse/SDK-723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Samples~/AvatarCreatorSamples/AvatarCreatorWizard/Scripts/UI/ReloadPreviouslyEditedAvatar.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    public class ReloadPreviouslyEditedAvatar : MonoBehaviour
+    {
+        private const string TAG = nameof(ReloadPreviouslyEditedAvatar);
+
+        [SerializeField] private Button continueEditingAvatar;
+        [SerializeField] private Button cancelEditingAvatar;
+
+        private string avatarId;
+        public UnityEvent<string> onSaveAvatar;
```
HarrisonHough, line 18: Rider warns about using capital letters for event names however I couldn't have it auto-enforce this convention. But it should be a capital first letter for public Unity events or actions or properties to keep it consistent across our SDK. 
```suggestion
        public UnityEvent<string> OnSaveAvatar;
```
Also I think this event should be renamed to something more like OnContinueEditingAvatar or something

Samples~/AvatarCreatorSamples/AvatarCreatorWizard/Scripts/UI/ReloadPreviouslyEditedAvatar.cs
```diff
@@ -0,0 +1,63 @@
...
+        public UnityEvent onCancel;
```
HarrisonHough, line 19:
```suggestion
        public UnityEvent OnCancel;
```

Samples~/AvatarCreatorSamples/AvatarCreatorWizard/Scripts/UI/ReloadPreviouslyEditedAvatar.cs
```diff
@@ -0,0 +1,63 @@
...
+                OnCancelEditing();
+                SDKLogger.Log(TAG, $"Failed to load draft avatar with id ${avatarId}. {e.Message}");
+            }
+        }
+        private void OnEnable()
+        {
+            continueEditingAvatar.onClick.AddListener(OnContinueEditing);
+            cancelEditingAvatar.onClick.AddListener(OnCancelEditing);
+        }
+        private void OnDisable()
```
HarrisonHough, line 39: You can run Riders code reformat and it will fix the spacing between functions. The functions should have 1 space in between each.

Samples~/AvatarCreatorSamples/AvatarCreatorWizard/Scripts/UI/ReloadPreviouslyEditedAvatar.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        private void ClearAvatarId()
+        {
+            AuthManager.StoreLastModifiedAvatar(null);
+            avatarId = null;
+        }
+
+
+    }
+}
```
HarrisonHough, line 63: Also an empty/new line should be at the end of every C# script. While it probably won't cause any issues, it's generally good practice and can prevent compiler or interpretor warnings. Also its good just for consistency.

HarrisonHough (changes requested):
Some minor naming and code formatting issues. 
It works fine however I think it would make more sense to have a LoadPreviousAvatarPopup element or something. So it could be used not only in Wizard Cac but in any CAC including our elements example.

rk132:
> Some minor naming and code formatting issues. It works fine however I think it would make more sense to have a LoadPreviousAvatarPopup element or something. So it could be used not only in Wizard Cac but in any CAC including our elements example.

I wouldn't add it as it is again a really specific case. But up for discussion.

HarrisonHough (approved):
Approved. 
However I do think that reloading previous avatar should be a core element in the avatar creator. Basically all character creator systems have the functionality that you can open the character creator and pick up where you left off. For RPM we could do this for those that are signed in.

PR #230, opened 2024-02-15
[SDK-820] - Update RPM account creation view and add it to sample scene before saving avatar

rk132:
## [SDK-820](https://ready-player-me.atlassian.net/browse/SDK-820)

image

## Description

-   Update save avatar to show log into RPM view

## How to Test

-   Test in Avatar Creator Elements scene

[SDK-820]: https://ready-player-me.atlassian.net/browse/SDK-820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HarrisonHough (changes requested):
Not related to this PR but we should remove the use of Update for checking the email field text.
It would be more efficient to make use of the Input field events like 
emailField.onValueChanged and/or emailField.onEndEdit as it will only check when things change (in this case we should also check OnEnable too just to be safe) 
Also perhaps it makes sense to change this to SignUpElement to be more consistent?

Samples~/AvatarCreatorSamples/AvatarCreatorElements/Scripts/SimpleAvatarCreator.cs
```diff
@@ -62,6 +63,16 @@ public async void LoadAvatar(string avatarId)
         }
 
         public async void SaveAvatar()
```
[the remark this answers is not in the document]
rk132: yeah, makes sense. I was thinking about it as well and couldn't figure anything good 😄

Runtime/AvatarCreator/Scripts/UI/Elements/AccountCreationElement.cs
```diff
@@ -20,27 +20,29 @@ public class AccountCreationElement : MonoBehaviour
...
 
-        private void Update()
-        {
-            var email = emailField.text;
-            sendEmailButton.interactable = !string.IsNullOrEmpty(email) && ValidatorUtil.IsValidEmail(email);
-        }
-
         private void OnEnable()
         {
+            emailField.onValueChanged.AddListener(OnEmailChanged);
```
HarrisonHough, line 25: Only thing is I think we should add this extra check just to be safe that when it is reenabled it will be non-interactable if empty. 
```suggestion
            var email = emailField.text;
            sendEmailButton.interactable = !string.IsNullOrEmpty(email) && ValidatorUtil.IsValidEmail(email);
            emailField.onValueChanged.AddListener(OnEmailChanged);
```

rk132: I made the button not interractable at the beginning. That should cover the cases

HarrisonHough (approved):
All good, nice work

PR #241, opened 2024-02-27
Mesh Helper multi-mesh support

srcnalt:
## Description

-   Mesh helper now supports transfer from multiple source mesh, material and bones target

## How to Test

-   Load atlassed and non-atlassed avatars with different customization combinations in Photon Sample project

## Checklist

-   [x] Changelog is updated.

[SDK-524]: https://ready-player-me.atlassian.net/browse/SDK-524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/Core/Scripts/Utils/AvatarMeshHelper.cs
+++ b/Runtime/Core/Scripts/Utils/AvatarMeshHelper.cs
@@ -5,6 +5,12 @@ namespace ReadyPlayerMe.Core
 {
     public static class AvatarMeshHelper
     {
+        [Obsolete("Use TransferMesh(GameObject source, GameObject target) instead.")]
+        public static void TransferMesh(GameObject source, SkinnedMeshRenderer[] targetMeshes, Animator targetAnimator)
+        {
+            TransferMesh(source, targetMeshes[0].transform.parent.gameObject);
+        }
+        
         /// <summary>
         ///     Transfers the mesh and material data from the source to the target avatar.
         /// </summary>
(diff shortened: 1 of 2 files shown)


Runtime/Core/Scripts/Utils/AvatarMeshHelper.cs
```diff
@@ -1,29 +1,63 @@
...
 {
     public static class AvatarMeshHelper
     {
-        public static void TransferMesh(GameObject source, SkinnedMeshRenderer[] targetMeshes, Animator targetAnimator)
+        /// <summary>
+        ///     Transfers the mesh and material data from the source to the target avatar.
+        /// </summary>
+        /// <param name="source">Imported Ready Player Me avatar.</param>
+        /// <param name="target">Photon Network Player Character.</param>
+        public static void TransferMesh(GameObject source, GameObject target)
```
rk132, line 13: This is breaking change?

srcnalt: Yes changed the signature.

rk132: What I meant is that this would require a move to the 7.0.0 version, since it might break functionality for the users. Do we want that?

srcnalt: This is a minor functionality change probably only used by people who used latest version of photon demo with this method, I would not expect it to be that impactful but you would know better if it is widely used somewhere.

rk132: I don't on usage, but I rather don't want to find it out 😅

rk132 (approved):
Looks good in general, but we need the old definition back as well.

Runtime/Core/Scripts/Utils/AvatarMeshHelper.cs
srcnalt: Okay now it wont break, I brought old signature, incase anyone still relied on it. Will work and warn that the method is obsolete.

PR #242, opened 2024-02-27
GetHeadMeshes method

srcnalt:
## Description

- GetHeadMeshes method helps developers get the list of game objects that are related to avatars head, which can be useful for hiding these in certain situations.
- Removed logic for removing unused meshes to avoid missing customizations upon next use.

## How to Test

-   Load a non-atlassed avatar and call the method avatar as it's only parameter.

## Checklist

-   [x] Changelog is updated.


rk132 (approved):
Looks good

PR #252, opened 2024-03-26
[SDK-800] Add body shapes

rk132:
## [SDK-800](https://ready-player-me.atlassian.net/browse/SDK-800)

image

## Description

- Added body shapes to the SDK
- Updated Elements scene to have the bodyshapes

## How to Test

- Check, that avatar changes

## Checklist

-   [ ] Changelog is updated.

[SDK-800]: https://ready-player-me.atlassian.net/browse/SDK-800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Runtime/AvatarCreator/Scripts/UI/Elements/BodyShapeSelectionElement.cs
```diff
@@ -0,0 +1,68 @@
...
+            availableBodyshapes = availableBodyshapes.Select(bodyShape => new AssetBodyShape
+            {
+                bodyShape = bodyShape.bodyShape,
+                AssetType = bodyShape.AssetType,
+                image = bodyShape.image,
+                Id = GetEnumDescription(bodyShape.bodyShape)
+            }).ToArray();
+        }
+
+        private static string GetEnumDescription(Enum value)
```
HarrisonHough, line 54: I think that it would be better to use an extension or helper method for this EG
BodyShape.GetAsID (perhaps you can think of a better name). 
```cs
public static class EnumExtensions
{
    public static string GetAsID(this Enum enumValue)
    {
        return enumValue.ToString().ToLower();
    }
}
```
The other thing is I wonder if it's not just easier (and perhaps slightly more optimized) to just get the id by something like 
```cs
BodyShape shape = BodyShape.Athletic;
string shapeName = shape.ToString().ToLower();
// shapeName = athletic
```
Instead of all this other stuff

HarrisonHough (changes requested):
Some comments and suggestions.
But otherwise it is working well

Samples~/AvatarCreatorSamples/AvatarCreatorElements/Scripts/SimpleAvatarCreator.cs
```diff
@@ -86,8 +72,58 @@ public async void SaveAvatar()
...
+            loading.SetActive(true);
+            var templateAvatarProps = await avatarManager.CreateAvatarFromTemplate(templateId, bodyType);
+
+            // Destroy the old avatar and replace it with the new one.
+            if (avatar != null)
+            {
+                Destroy(avatar);
+            }
+            var previousGender = gender;
+            avatar = templateAvatarProps.Item1;
```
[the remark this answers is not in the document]
HarrisonHough: Some IDE's show return types but not all.  
It looks clear for us because we know what they are. 
If you are trying to understand the code from a Git PR for example it is very difficult :D 
If it was something hidden from developers it wouldn't be a problem but our public API's return the tuples.

Samples~/AvatarCreatorSamples/AvatarCreatorElements/Scripts/ElementSectionManager.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace ReadyPlayerMe.Samples.AvatarCreatorElements
+{
+    public class ElementSectionManager : MonoBehaviour
```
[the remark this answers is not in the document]
rk132: did some renaming there.

Runtime/AvatarCreator/Scripts/UI/Elements/BodyShapeSelectionElement.cs
rk132: Hmm, not a fan of messing with IDs (upperToLower) as they should always be the same. And the description is also probably the wrong place for it. Or create a customProperty called Id and add it?

``` C#
using System;

[AttributeUsage(AttributeTargets.Field, AllowMultiple = false)]
public class IdAttribute : Attribute
{
    public string Id { get; private set; }

    public IdAttribute(string id)
    {
        Id = id;
    }
}
```

``` C#
public enum BodyShape
{
    None,
    [Id("average")] Average,
    [Id("athletic")] Athletic,
    [Id("heavyset")] HeavySet,
    [Id("plussize")] PlusSize
}
```

using System.Reflection;

```C#
public static class EnumExtensions
{
    public static string GetId(this Enum value)
    {
        var field = value.GetType().GetField(value.ToString());
        var attribute = Attribute.GetCustomAttribute(field, typeof(IdAttribute)) as IdAttribute;
        return attribute == null ? value.ToString() : attribute.Id;
    }
}
```

Samples~/AvatarCreatorSamples/AvatarCreatorElements/Scripts/SimpleAvatarCreator.cs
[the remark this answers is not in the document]
rk132: Hmm, yeah that makes sense. I'll create a task for it.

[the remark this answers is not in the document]
rk132: But it's breaking change, so we need a new version bump at some point for this.

Runtime/AvatarCreator/Scripts/UI/Elements/BodyShapeSelectionElement.cs
HarrisonHough: Don't you think it's overcomplicating it? 
Also we use the same thing in other places 
image

rk132: Example there would be the following: [Description("fullbody-xr")] FullBodyXR. This wouldn't work, since the id should be fullbody-xr. So it works for this case, but it is prone to error.

rk132: Or we have this kind of methods
https://github.com/readyplayerme/rpm-unity-sdk-core/blob/main/Runtime/AvatarCreator/Scripts/WebRequests/AvatarAPIRequests.cs#L98-L109

HarrisonHough: ok just leave them how they are we can review later

rk132: Also regarding efficiency it's about twice as slow, but one call takes 0.00059 seconds, so not too worried about that.

HarrisonHough (approved):

PR #268, opened 2024-05-06
[SDK-934] - initialize request in property

rk132:
## Description

-   Remove race condition, when assetSelectionElement is disabled, but it is initialized before it is enabled.

## How to Test

-   When assetSelectionElement is disabled, you can still load the buttons.

[SDK-934]: https://ready-player-me.atlassian.net/browse/SDK-934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/AvatarCreator/Scripts/UI/Elements/AssetSelectionElement.cs
+++ b/Runtime/AvatarCreator/Scripts/UI/Elements/AssetSelectionElement.cs
@@ -21,11 +21,12 @@ namespace ReadyPlayerMe.AvatarCreator
         [SerializeField] private int iconSize = 64;
 
         private PartnerAsset[] assets;
+        
         private AssetAPIRequests assetsRequests;
+        private AssetAPIRequests AssetsRequests => assetsRequests ??= new AssetAPIRequests(CoreSettingsHandler.CoreSettings.AppId);
 
         private void Awake()
         {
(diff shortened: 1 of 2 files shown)


Runtime/AvatarCreator/Scripts/UI/Elements/AssetSelectionElement.cs
```diff
@@ -21,11 +21,12 @@ public class AssetSelectionElement : SelectionElement
         [SerializeField] private int iconSize = 64;
 
         private PartnerAsset[] assets;
+        
         private AssetAPIRequests assetsRequests;
+        private AssetAPIRequests AssetsRequests => assetsRequests ??= new AssetAPIRequests(CoreSettingsHandler.CoreSettings.AppId);
```
HarrisonHough, line 26: Why not just use this ? 
private AssetAPIRequests assetsRequests = new AssetAPIRequests(CoreSettingsHandler.CoreSettings.AppId); 
Is the extra "AssetRequests" property needed? It's also private so not sure it makes sense to keep it there?

HarrisonHough (approved):
looks good, nice catch with the blendshapes thing

PR #270, opened 2024-05-10
[SDK-943] Feature/template filter

HarrisonHough:
## Description

-   current release gets all templates and they don't match well together, it also bloats the templates panel a bit making it take longer 
image
this PR includes the ability to filter these, it still requests all of them (because the endpoint param doesn't work) but then filters before downloading all the renders. 
- it's now exposed in the TemplateSelectionElement 
image
I also renamed the options to be V1, V2 because I feel like the naming on the backend is not really relevant to the use ("onboarding", "randomize") Onboarding are the new ones so I named them V2. 

## How to Test

-   open elements creator or a scene with Template selection prefab and adjust the TemplateVersions property and run the scene to test. 

## Checklist

-   [ ] Changelog is updated.

### DIFF
--- a/Runtime/AvatarCreator/Scripts/AvatarTemplateFetcher.cs
+++ b/Runtime/AvatarCreator/Scripts/AvatarTemplateFetcher.cs
@@ -7,6 +7,13 @@ using ReadyPlayerMe.Core;
 
 namespace ReadyPlayerMe.AvatarCreator
 {
+    public enum TemplateUsageType
+    {
+        All,
+        Onboarding,
+        Randomize
+    }
+
     /// <summary>
     /// This class can be used to fetch avatar template data including icon renders from the avatarAPI.
     /// </summary>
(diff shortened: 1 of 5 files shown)


Runtime/AvatarCreator/Scripts/WebRequests/AvatarAPIRequests.cs
```diff
@@ -24,6 +24,7 @@ public class AvatarAPIRequests
         private const string PARTNER = "partner";
         private const string DATA = "data";
         private const string ID = "id";
+        private const string TEMPLATE_ONBOARDING_USAGE_TYPE = "onboarding";
```
rk132, line 27: Not needed

Runtime/AvatarCreator/Scripts/UI/Elements/TemplateSelectionElement.cs
```diff
@@ -1,6 +1,9 @@
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using ReadyPlayerMe.Core;
+using UnityEngine;
+using UnityEngine.Serialization;
```
rk132, line 6: not needed

Runtime/AvatarCreator/Scripts/AvatarTemplateFetcher.cs
```diff
@@ -7,17 +7,28 @@
...
+
     /// <summary>
     /// This class can be used to fetch avatar template data including icon renders from the avatarAPI.
     /// </summary>
     public class AvatarTemplateFetcher
     {
         private readonly CancellationToken ctx;
         private readonly AvatarAPIRequests avatarAPIRequests;
+        private const string TEMPLATE_ONBOARDING_USAGE_TYPE = "onboarding";
+        private const string TEMPLATE_RANDOMIZE_USAGE_TYPE = "randomize";
```
rk132, line 25: I dont understand the naming of it, but would it make sense to rename the variables to 

TEMPLATE_V1_USAGE_TYPE and TEMPLATE_V2_USAGE_TYPE. Onboarding and Randomize don't say much to be fair :D .

rk132 (approved):
Looks good in general

Runtime/AvatarCreator/Scripts/WebRequests/AvatarAPIRequests.cs
HarrisonHough: updated

PR #272, opened 2024-05-14
feat: update icons

rk132:
## Description

-   Update icons

[SDK-805]: https://ready-player-me.atlassian.net/browse/SDK-805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HarrisonHough (approved):
Looks good. 
We have multiple copies of hairstyle, headwear & outfit icons. Maybe we can make a separate folder EG "SharedAssets" in the root AvatarCreatorSamples folder to remove this duplication. 
 
Another thing we could also do is make a texture atlas from the icons and just have them all in 1 single texture. 
Main benefit would be 1 single file, but I guess the downside is if you need to update or add more it could be annoying to maintain 😅

rk132:
> Looks good. We have multiple copies of hairstyle, headwear & outfit icons. Maybe we can make a separate folder EG "SharedAssets" in the root AvatarCreatorSamples folder to remove this duplication.
> 
> Another thing we could also do is make a texture atlas from the icons and just have them all in 1 single texture. Main benefit would be 1 single file, but I guess the downside is if you need to update or add more it could be annoying to maintain 😅

yeah, but the images are actually sample-specific, even tho, they are the same. I wouldn't move them under the core for now.

PR #273, opened 2024-05-14
feat: add gender option to template selection element

rk132:
## Description

-   Add option to filter template elements by gender

## How to Test

-   Check, that filtering works for gender

[SDK-797]: https://ready-player-me.atlassian.net/browse/SDK-797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/AvatarCreator/Scripts/AvatarTemplateFetcher.cs
+++ b/Runtime/AvatarCreator/Scripts/AvatarTemplateFetcher.cs
@@ -21,14 +21,10 @@ namespace ReadyPlayerMe.AvatarCreator
     {
         private readonly CancellationToken ctx;
         private readonly AvatarAPIRequests avatarAPIRequests;
-        private const string TEMPLATE_V2_USAGE_TYPE = "onboarding";
-        private const string TEMPLATE_V1_USAGE_TYPE = "randomize";
-        private readonly TemplateVersions templateVersions;
 
-        public AvatarTemplateFetcher(CancellationToken ctx = default, TemplateVersions templateVersions = TemplateVersions.V2)
+        public AvatarTemplateFetcher(CancellationToken ctx = default)
         {
             this.ctx = ctx;
-            this.templateVersions = templateVersions;
             avatarAPIRequests = new AvatarAPIRequests(ctx);
         }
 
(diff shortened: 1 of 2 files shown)


Runtime/AvatarCreator/Scripts/AvatarTemplateFetcher.cs
```diff
@@ -36,19 +36,32 @@ public AvatarTemplateFetcher(CancellationToken ctx = default, TemplateVersions t
         /// Fetches all avatar templates without the icon renders via the avatarAPI.
         /// </summary>
         /// <returns></returns>
-        public async Task<List<AvatarTemplateData>> GetTemplates()
+        public async Task<List<AvatarTemplateData>> GetTemplates(OutfitGender gender = OutfitGender.None)
         {
             var templates = await avatarAPIRequests.GetAvatarTemplates();
-            switch (templateVersions)
+            return templates.Where(template => HasCorrectTemplateVersion(template) && HasCorrectGender(template)).ToList();
```
HarrisonHough, line 42: With the gender filtering I was mostly thinking about being able to hide/show only male/female template avatar icons. I think it still makes sense to still download both male and female data + renders as it is very unlikely an app would only use 1 gender, so you may as well fetch both outfit genders at once initially and then in the SelectionElement you will then just set the visibility based on which gender you want to see. 

Also this now makes me thing that probably this is also the wrong place to be filtering the "Template version". Because the request for data (no renders) is super light weight and fast. So in that sense maybe it also makes sense to delegate the Template version filtering to the TemplateSelectionElement so that you just filter the ones you don't want before requesting the renders, if that makes sense.

HarrisonHough (approved):
No blockers but I think there might be a more practical way to do the gender filtering. Happy to discuss if you want.

Runtime/AvatarCreator/Scripts/AvatarTemplateFetcher.cs
rk132: yep, agree with this.

Runtime/AvatarCreator/Scripts/UI/Elements/TemplateSelectionElement.cs
```diff
@@ -31,15 +36,52 @@ private void Awake()
...
+            
+            CreateButtons(filteredTemplates!.ToArray(), async (button, asset) =>
             {
                 var webRequestDispatcher = new WebRequestDispatcher();
                 var url = $"{asset.ImageUrl}";
                 var texture = await webRequestDispatcher.DownloadTexture(url);
                 button.SetIcon(texture);
             });
         }
+        private bool HasCorrectTemplateVersion(AvatarTemplateData template)
```
HarrisonHough, line 63: please add empty line in between functions

Runtime/AvatarCreator/Scripts/UI/Elements/TemplateSelectionElement.cs
```diff
@@ -31,15 +36,52 @@ private void Awake()
...
+        /// Each button is created with an icon fetched from the template's image URL.
+        /// </summary>
+        public async Task LoadAndCreateButtons(bool useCachedResponse)
+        {
+            if (!useCachedResponse || avatarTemplates == null)
+            {
+                await LoadTemplateData();
+            }
+
+            var filteredTemplates = avatarTemplates!.Where(template => HasCorrectTemplateVersion(template) && HasCorrectGender(template)).ToList();
```
HarrisonHough, line 53: I would still think we should remove the gender filter out of here. So then it will always create all the gendered template buttons, and then you also wouldn't need the useCachedResponse logic there. 
Instead add a line inside the CreateButtons function callback to set the visibility button visibility based on the gender. 
EG
```cs
                button.SetIcon(texture);
                button.gameObject.SetActive(asset.Gender == gender);
```

We should also then probably make a function that enables you to switch between male only, female only and all gender outfits. 
```cs
        public void SetVisibilityFilter(OutfitGender gender)
        {
            this.gender = gender;
            foreach (var avatarTemplate in avatarTemplates)
            {
                var button = GetButton(avatarTemplate.Id);
                if(button == null) continue;
                button.gameObject.SetActive(HasCorrectGender(avatarTemplate));
            }
        }
```

HarrisonHough (approved):
All good just an idea slightly different approach.

PR #277, opened 2024-05-20
fix: check if the lod file update time is creater than the updated av…

rk132:
## Description

-   issues with LOD's and avatars not being updated, when cache is enabled.

## How to Test

- Create an avatar online with any application
- load avatar with LOD sample (paste avatar URL and load for the first time)
- update same avatar in the web
- verify, that the avatar is updated for all of the LODs

[SDK-718]: https://ready-player-me.atlassian.net/browse/SDK-718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/Core/Scripts/Operations/MetadataDownloader.cs
+++ b/Runtime/Core/Scripts/Operations/MetadataDownloader.cs
@@ -129,7 +129,17 @@ namespace ReadyPlayerMe.Core
                 return true;
             }
             AvatarMetadata previousMetadata = AvatarMetadata.LoadFromFile(context.AvatarUri.LocalMetadataPath);
-            return AvatarMetadata.IsUpdated(context.Metadata, previousMetadata);
+            return AvatarMetadata.IsUpdated(context.Metadata, previousMetadata) || ShouldUpdateAvatarFile(context);
+        }
+
+        private static bool ShouldUpdateAvatarFile(AvatarContext context)
+        {
+            if (!File.Exists(context.AvatarUri.LocalModelPath))
+            {
+                return true;
+            }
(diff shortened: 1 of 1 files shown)


Runtime/Core/Scripts/Operations/MetadataDownloader.cs
```diff
@@ -129,7 +129,17 @@ private static bool IsUpdateRequired(AvatarContext context)
...
+            return AvatarMetadata.IsUpdated(context.Metadata, previousMetadata) || ShouldUpdateAvatarFile(context);
+        }
+
+        private static bool ShouldUpdateAvatarFile(AvatarContext context)
+        {
+            if (!File.Exists(context.AvatarUri.LocalModelPath))
+            {
+                return true;
+            }
+            var avatarFileUpdateTime = File.GetLastWriteTime(context.AvatarUri.LocalModelPath);
```
HarrisonHough, line 141: I think it is better to use UTC to prevent potential issues with local time. 
```cs
            var avatarFileUpdateTime = File.GetLastWriteTimeUtc(context.AvatarUri.LocalModelPath);
            return avatarFileUpdateTime < context.Metadata.UpdatedAt;
```

HarrisonHough (approved):
It works great idea. 
I do think there might be better ways to do it like reworking LOD or caching for LODS, and I'm not sure how reliable File.GetLastWriteTime is or if there are issues with different platforms (EG mobile). But if we do lots of testing on this and it all works then all good.

PR #281, opened 2024-05-24
feat: log error, when bodyshapes are not enabled in studio

rk132:
## Description

-   Throw error, when body shapes are not enabled in the studio
image

## How to Test

- Enable body shapes in the studio and test, that shapes are applied and that no error is thrown
- Disable body shapes in the studio and verify, that error is shown and avatar is not updated

## Checklist

-   [ ] Changelog is updated.

[SDK-907]: https://ready-player-me.atlassian.net/browse/SDK-907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
+++ b/Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
@@ -25,6 +25,8 @@ namespace ReadyPlayerMe.AvatarCreator
     /// </summary>
     public class AvatarManager : IDisposable
     {
+        private const string ERROR_BODYSHAPES_NOT_ENABLED = "Failed to apply body shapes to the avatar. Please ensure body shapes are enabled in your studio application.";
+        
         private const string TAG = nameof(AvatarManager);
         private readonly AvatarAPIRequests avatarAPIRequests;
         private readonly string avatarConfigParameters;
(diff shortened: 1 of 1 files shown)


Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
```diff
@@ -288,9 +290,32 @@ public async Task<GameObject> UpdateAsset(AssetType assetType, BodyType bodyType
             {
                 return null;
             }
+            
+#if UNITY_EDITOR
```
HarrisonHough, line 294: Why did you make it editor only?

Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
```diff
@@ -288,9 +290,32 @@ public async Task<GameObject> UpdateAsset(AssetType assetType, BodyType bodyType
...
+            await ValidateAssetUpdate(assetType, assetId);
+#endif
 
             return await inCreatorAvatarLoader.Load(avatarId, bodyType, gender, data);
         }
+        
+#if UNITY_EDITOR
+        private async Task ValidateAssetUpdate(AssetType assetType, object assetId)
+        {
+            if (assetType != AssetType.BodyShape)
```
HarrisonHough, line 304: because of this check here, probably you should rename it to something like ValidateBodyShapes or CheckForBodyshapes.
However if you want to keep it as a more generic Validate asset check that can be used on any asset type, then I would remove this if statement and instead wrap the calling of this function in it.  
EG  
```cs
if(assetType == AssetType.BodyShape){
await ValidateAssetUpdate(assetType, assetId);
}
```

Although if you do this you might need to make the error message more dynamic, like "failed to apply {assetType}..." etc.

HarrisonHough (commented):
Some questions about implementation

Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
rk132: No point in showing this error in live environment (This makes extra requests each time)

rk132: Maybe, if we do asset validation in the future, where we get the correct response immediately, then we can think about it

Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
rk132: good point

Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
HarrisonHough: Hmm disagree. 
If somebody tries to use bodyshapes in builds, it should log an error if it's not possible to use them. 
I realize that they would see this when they test in editor but it's still entirely possible that they might just straight cut and paste our elements example and then build and wonder why it's not working.  
It is an extra request but it only happens when you try to use the body shapes element, if they don't want the extra requests then they should remove it.

HarrisonHough (approved):
It's okay how it is but I would still go with logging errors on builds also. It makes it easier to debug things if somebody misses the error when testing in editor.

PR #285, opened 2024-06-03
Feature/add cancellation tokens

rk132:
## Description

-   Update methods to throw an error, when cancellationToken was called, and handle them in the Elements scene

[SDK-971]: https://ready-player-me.atlassian.net/browse/SDK-971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/AvatarCreator/Scripts/Extensions/TaskExtensions.cs
+++ b/Runtime/AvatarCreator/Scripts/Extensions/TaskExtensions.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Threading.Tasks;
+
+namespace ReadyPlayerMe.AvatarCreator
+{
+    public static class TaskExtensions
+    {
+        public const string ON_REQUEST_CANCELLED_MESSAGE = "Request was cancelled"; 
+        public static async Task HandleCancellation(Task taskToWait, Action onSuccess = null)
+        {
+            try
+            {
+                await taskToWait;
+                onSuccess?.Invoke();
+            }
+            catch (Exception ex)
+            {
+                if (ex.Message != ON_REQUEST_CANCELLED_MESSAGE)
+                {
+                    throw;
+                }
+            }
+        }
+        public static async Task<T> HandleCancellation<T>(Task<T> taskToWait)
+        {
+            try
+            {
+                return await taskToWait;
+            }
+            catch (Exception ex)
+            {
+                if (ex.Message != ON_REQUEST_CANCELLED_MESSAGE)
+                {
+                    throw;
+                }
+            }
+            return default;
+        }
+    }
+}
--- a/Runtime/AvatarCreator/Scripts/Extensions/TaskExtensions.cs
+++ b/Runtime/AvatarCreator/Scripts/Extensions/TaskExtensions.cs
@@ -5,7 +5,8 @@ namespace ReadyPlayerMe.AvatarCreator
 {
     public static class TaskExtensions
     {
-        public const string ON_REQUEST_CANCELLED_MESSAGE = "Request was cancelled"; 
+        public const string ON_REQUEST_CANCELLED_MESSAGE = "Request was cancelled";
+        private const string ON_OPERATION_CANCELLED_MESSAGE = "Operation was cancelled";
         public static async Task HandleCancellation(Task taskToWait, Action onSuccess = null)
         {
             try
@@ -15,12 +16,13 @@ namespace ReadyPlayerMe.AvatarCreator
             }
             catch (Exception ex)
             {
-                if (ex.Message != ON_REQUEST_CANCELLED_MESSAGE)
+                if (ex.Message != ON_REQUEST_CANCELLED_MESSAGE && ex.Message != ON_OPERATION_CANCELLED_MESSAGE)
                 {
                     throw;
                 }
             }
         }
+        
         public static async Task<T> HandleCancellation<T>(Task<T> taskToWait)
         {
             try
@@ -29,7 +31,7 @@ namespace ReadyPlayerMe.AvatarCreator
             }
             catch (Exception ex)
             {
-                if (ex.Message != ON_REQUEST_CANCELLED_MESSAGE)
+                if (ex.Message != ON_REQUEST_CANCELLED_MESSAGE && ex.Message != ON_OPERATION_CANCELLED_MESSAGE)
                 {
                     throw;
                 }
--- a/Runtime/AvatarCreator/Scripts/Managers/AuthManager.cs
+++ b/Runtime/AvatarCreator/Scripts/Managers/AuthManager.cs
@@ -41,7 +41,7 @@ namespace ReadyPlayerMe.AvatarCreator
             OnSignedIn?.Invoke(userSession);
         }
 
-        public static async void SendEmailCode(string email, CancellationToken cancellationToken = default)
+        public static async Task SendEmailCode(string email, CancellationToken cancellationToken = default)
         {
             try
             {
@@ -69,7 +69,7 @@ namespace ReadyPlayerMe.AvatarCreator
             }
         }
 
-        public static async void Signup(string email, CancellationToken cancellationToken = default)
+        public static async Task Signup(string email, CancellationToken cancellationToken = default)
         {
             try
             {
--- a/Runtime/AvatarCreator/Scripts/Managers/AuthManager.cs
+++ b/Runtime/AvatarCreator/Scripts/Managers/AuthManager.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using ReadyPlayerMe.Core;
 
@@ -27,9 +28,9 @@ namespace ReadyPlayerMe.AvatarCreator
             AuthAPIRequests = new AuthAPIRequests(CoreSettingsHandler.CoreSettings.Subdomain);
         }
 
-        public static async Task LoginAsAnonymous()
+        public static async Task LoginAsAnonymous(CancellationToken cancellationToken = default)
         {
-            userSession = await AuthAPIRequests.LoginAsAnonymous();
+            userSession = await AuthAPIRequests.LoginAsAnonymous(cancellationToken);
             IsSignedInAnonymously = true;
         }
 
@@ -40,11 +41,11 @@ namespace ReadyPlayerMe.AvatarCreator
             OnSignedIn?.Invoke(userSession);
         }
 
-        public static async void SendEmailCode(string email)
+        public static async void SendEmailCode(string email, CancellationToken cancellationToken = default)
         {
             try
             {
-                await AuthAPIRequests.SendCodeToEmail(email, userSession.Id);
+                await AuthAPIRequests.SendCodeToEmail(email, userSession.Id, cancellationToken);
             }
             catch (Exception e)
             {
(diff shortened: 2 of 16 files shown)


Runtime/AvatarCreator/Scripts/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        public const string ON_REQUEST_CANCELLED_MESSAGE = "Request was cancelled"; 
+        public static async Task HandleCancellation(Task taskToWait, Action onSuccess = null)
+        {
+            try
+            {
+                await taskToWait;
+                onSuccess?.Invoke();
+            }
+            catch (Exception ex)
```
HarrisonHough, line 16: If you want to listen for cancellations you can do this
```suggestion
            catch (OperationCanceledException ex)
```
Then you don't need to check the message name and it will catch all cancellations

Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
```diff
@@ -76,6 +76,7 @@ public async Task<AvatarCreationResponse> CreateAvatarAsync(AvatarProperties ava
             }
             catch (Exception e)
             {
+                ThrowErrIfCancellationException(e);
```
HarrisonHough, line 79: It's super annoying that we have to add this in all the exception parts of this script. 
It could be moved into some sort of "HandleException(ex)" function to reduce the code duplication.
It might also be nice in future to change the OnError action to have a parameter of type Exception. Then we can just use that for everthing (including throwing the cancellation exception). It would be a breaking change though.

Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
```diff
@@ -373,5 +384,13 @@ public void Dispose()
         {
             ctxSource?.Cancel();
         }
+
+        private void ThrowErrIfCancellationException(Exception e)
+        {
+            if (e.Message == TaskExtensions.ON_REQUEST_CANCELLED_MESSAGE)
```
HarrisonHough, line 390: you can check for OperationCanceledException here instead of the string comparison

HarrisonHough (changes requested):
cancellation token stuff is okay but I don't think we should be throwing errors like this. 
people are going to freak out when they see all these errors even though it's not a breaking error
image
maybe warnings or something but even then it's gonna spam the logs, happy to discuss though

Runtime/AvatarCreator/Scripts/UI/Elements/LoginElement.cs
```diff
@@ -32,12 +34,18 @@ private void OnDisable()
...
+        }
+
         /// <summary>
         /// Sends a verification code to the email address that was entered into the email InputField.
         /// </summary>
-        public void SendVerificationCode()
+        public async void SendVerificationCode()
         {
-            AuthManager.SendEmailCode(emailField.text);
+            await TaskExtensions.HandleCancellation(AuthManager.SendEmailCode(emailField.text, cancellationTokenSource.Token));
```
HarrisonHough, line 48: I appreciate that you want to address all the asnyc events/requests but think this cancellation token is overkill. If this function did something destructive or that changed the state and took a significant time then it might make sense but its super fast whether this runs or not it doesn't really effect much.

Runtime/AvatarCreator/Scripts/Managers/AuthManager.cs
```diff
@@ -40,23 +41,23 @@ public static void SetUser(UserSession session)
             OnSignedIn?.Invoke(userSession);
         }
 
-        public static async void SendEmailCode(string email)
+        public static async Task SendEmailCode(string email, CancellationToken cancellationToken = default)
         {
             try
             {
-                await AuthAPIRequests.SendCodeToEmail(email, userSession.Id);
+                await AuthAPIRequests.SendCodeToEmail(email, userSession.Id, cancellationToken);
```
HarrisonHough, line 48: I think this cancel token is not necessary

Runtime/AvatarCreator/Scripts/UI/Elements/DeleteAvatarElement.cs
```diff
@@ -21,12 +22,20 @@ public class DeleteAvatarElement : MonoBehaviour
 
         private string avatarId;
         private AvatarAPIRequests avatarAPIRequests;
+        private CancellationTokenSource cancellationTokenSource;
```
HarrisonHough, line 25: why not just do this? 

private CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();

Runtime/AvatarCreator/Scripts/UI/Elements/LoginElement.cs
```diff
@@ -53,13 +61,17 @@ public async void LoginWithCode()
...
                 var userIdToMerge = mergeCurrentSession && AuthManager.IsSignedInAnonymously ? AuthManager.UserSession.Id : null;
-                if (await AuthManager.LoginWithCode(codeField.text, userIdToMerge))
+                if (await AuthManager.LoginWithCode(codeField.text, userIdToMerge, cancellationTokenSource.Token))
                 {
                     LoginSuccess();
                 }
             }
             catch (Exception e)
             {
+                if (e.Message == TaskExtensions.ON_REQUEST_CANCELLED_MESSAGE)
```
HarrisonHough, line 71: same as other message I think this isn't necessary?

Runtime/AvatarCreator/Scripts/UI/Elements/SignupElement.cs
```diff
@@ -34,15 +36,25 @@ private void OnDisable()
...
         {
             sendEmailButton.interactable = !string.IsNullOrEmpty(newEmailValue) && ValidatorUtil.IsValidEmail(newEmailValue);
         }
 
-        private void OnSendEmailButton()
+        private async void OnSendEmailButton()
         {
             var email = emailField.text;
-            AuthManager.Signup(email);
+            await TaskExtensions.HandleCancellation(AuthManager.Signup(email, cancellationTokenSource.Token), () => OnEmailSent(email));
```
HarrisonHough, line 53: same as before I'm not really sure why this HandleCancellation function is needed. The cancellation and exception throwing should be in the AuthManager.Signup or AuthAPIRequests.Signup. that way it will automatically be available for anybody using our public API's too

Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
rk132: response exceptions are regular exceptions, not OperationCanceledExceptions
https://github.com/readyplayerme/rpm-unity-sdk-core/blob/main/Runtime/AvatarCreator/Scripts/Extensions/ResponseExtensions.cs

Runtime/AvatarCreator/Scripts/UI/Elements/LogoutElement.cs
```diff
@@ -32,11 +34,15 @@ private async void Logout()
             AuthManager.Logout();
             try
             {
-                await AuthManager.LoginAsAnonymous();
+                await AuthManager.LoginAsAnonymous(cancellationTokenSource.Token);
                 LogOutSucceeded();
             }
             catch (Exception e)
```
HarrisonHough, line 40: you could also use this 
```suggestion
            catch (Exception ex) when (ex is not OperationCanceledException)
```

Runtime/AvatarCreator/Scripts/Extensions/TaskExtensions.cs
MaxAndreassenRPM: If the intention is to only catch cancellation token exceptions, and not catch anything else, then catching that specific error is generally the best way to go.

https://learn.microsoft.com/en-us/dotnet/standard/exceptions/how-to-use-specific-exceptions-in-a-catch-block

Runtime/AvatarCreator/Scripts/Managers/AvatarManager.cs
MaxAndreassenRPM: You can actually use the built in exclude logic to handle this, if you wanted a more graceful solution ->

`catch (Exception ex) when (ex is not OperationCanceledException)`

https://stackoverflow.com/questions/12633903/elegantly-handle-task-cancellation

Runtime/AvatarCreator/Scripts/UI/Elements/LoginElement.cs
rk132: There's nothing in the "Super fast" with network connections. I wanted to cover everything so that there are no issues with it. I understand, that it might be overkill, but still, it's a place, that might have a point of failure.

Runtime/AvatarCreator/Scripts/Extensions/TaskExtensions.cs
MaxAndreassenRPM: Ah, but I see, you're saying this is catching a middle layer of response exceptions that we've defined, so the actual type of the exception is lost before it reaches here?

Runtime/AvatarCreator/Scripts/Managers/AuthManager.cs
rk132: Right now, there's no token to pass separately to AuthAPI.

Runtime/AvatarCreator/Scripts/UI/Elements/ColorSelectionElement.cs
```diff
@@ -13,16 +15,33 @@ public class ColorSelectionElement : SelectionElement
...
+        private void OnDestroy()
+        {
+            cancellationTokenSource.Cancel();
+            cancellationTokenSource.Dispose();
+        }
+
         public async void LoadAndCreateButtons(AvatarProperties avatarProperties)
         {
-            await LoadColorPalette(avatarProperties);
+            await TaskExtensions.HandleCancellation(LoadColorPalette(avatarProperties));
```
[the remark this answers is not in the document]
rk132: The token is already passed to the avatarApi in constructor. It's just the point, where we want to catch the exception, that the call was cancelled and we should drop the call. That why there is a null check as well.

Runtime/AvatarCreator/Scripts/UI/Elements/DeleteAvatarElement.cs
rk132: good point 👍

Runtime/AvatarCreator/Scripts/UI/Elements/DeleteAvatarElement.cs
```diff
@@ -55,6 +64,10 @@ private async void DeleteAvatar()
             }
             catch (Exception e)
             {
+                if (e.Message == TaskExtensions.ON_REQUEST_CANCELLED_MESSAGE)
```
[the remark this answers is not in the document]
rk132: yep, true

Runtime/AvatarCreator/Scripts/UI/Elements/SignupElement.cs
rk132: It's not throwing an exception, but catching it here so that it wouldn't fall to the next level and throw unhandled exceptions. Elements are the last layer, that should handle the exceptions, that is not called out from next level (Creator controllers, that create buttons, etc..)

Runtime/AvatarCreator/Scripts/UI/Elements/LogoutElement.cs
rk132: It's not OperationCanceledException for requests

rk132: response exceptions are regular exceptions, not OperationCanceledExceptions
https://github.com/readyplayerme/rpm-unity-sdk-core/blob/main/Runtime/AvatarCreator/Scripts/Extensions/ResponseExtensions.cs

Runtime/AvatarCreator/Scripts/Extensions/TaskExtensions.cs
rk132: yep

HarrisonHough (approved):
Great work on this btw

PR #289, opened 2024-06-06
feat: refactor avatar creator scene

rk132:
## Description

-   Refactor avatar creator elements scene and separate avatar management from simple avatar creator script

[SDK-968]: https://ready-player-me.atlassian.net/browse/SDK-968?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HarrisonHough (approved):
Seems to work. 
I still don't like that we have to wrap everything in this TaskExtensions.HandleCancellation it makes the code messy and harder to read. 
At some point I would like to explore alternative options, perhaps a big refactor (or re-write) of how the core requests work so that we can handle this more cleanly. But it's fine for now

PR #290, opened 2024-06-11
[SDK-984] - move bodytype to settings

rk132:
## [SDK-984](https://ready-player-me.atlassian.net/browse/SDK-984)

image

## Description

-   Bodytype is moved to the settings page and removed from configurations. This means, that users don't need to set it manually everywhere.

[SDK-984]: https://ready-player-me.atlassian.net/browse/SDK-984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Runtime/AvatarCreator/Scripts/WebRequests/Responses/GetUserAvatarsResponse.cs
```diff
@@ -0,0 +1,11 @@
+using ReadyPlayerMe.Core;
+
+namespace ReadyPlayerMe.AvatarCreator.Responses
+{
+    public struct GetUserAvatarResponse
```
HarrisonHough, line 5: is the "Get" really necessary for this struct type name? 
It seems a little misleading as you usually use Get for a function that retrieves a value. 
EG GetAvatarID. 
I would say for a struct or type name UserAvatarResponse would be better.

Samples~/AvatarCreatorSamples/AvatarCreatorWizard/Scripts/UI/CategoryUICreator.cs
```diff
@@ -169,28 +168,31 @@ private void SelectCategoryGroup(AssetType category)
...
-                    cameraZoom.ToFaceView();
-                }
+                return;
+            }
+            
+            if (category.IsOutfitAsset())
+            {
+                cameraZoom.ToFullbodyView();
+            }
+            else
```
HarrisonHough, line 193: the else is kinda redundant you an just add a return after 
cameraZoom.ToFullbodyView();

HarrisonHough (approved):
Small stuff in comments. 
Overall looks good, it's a lot of changes but I like that it reduces the number of parameters required in a lot of functions.

PR #293, opened 2024-06-14
fix: allow cache to be loaded from previous versions, where bodyType was stored as an integer

rk132:
## Description

-   Previous versions stored the body type as an integer. This is fixed, so that they can be now loaded as an integer.

## How to Test

-   Save the avatar in the cache in version 5.0.0 for example, then update SDK to the latest and verify, that the avatars can be still loaded.


HarrisonHough (approved):
Looks good!

PR #295, opened 2024-06-17
[SDK-999] Fix mesh compression settings

HarrisonHough:
## Description

-   This PR addresses the folllowing issues 
    - the two settings are not compatible with each other. EG if 1 is enabled then the other can not be (it causes issues) So we should make it so that if 1 is enabled the other is automatically set to false (if it isn’t already)
    - Currently they can’t be toggled off. there seems to be an issue that once the plugins are installed they can’t be disabled. we need to fix that 

## How to Test

## Checklist

-   [ ] Changelog is updated.


rk132 (approved):
Looks good, but merge to the hotfix branch instead of develop

PR #296, opened 2024-06-17
[SDK-980] fetch bodytype specific templates

rk132:
## Description

-   Fetch body type specific templates

## How to Test

-   Verify, that correct templates are selected

[SDK-980]: https://ready-player-me.atlassian.net/browse/SDK-980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/AvatarCreator/Scripts/WebRequests/AvatarAPIRequests.cs
+++ b/Runtime/AvatarCreator/Scripts/WebRequests/AvatarAPIRequests.cs
@@ -20,9 +20,6 @@ namespace ReadyPlayerMe.AvatarCreator
         private const string RPM_AVATAR_V2_BASE_URL = Env.RPM_API_V2_BASE_URL + "avatars";
 
         private const string COLOR_PARAMETERS = "colors?type=skin,beard,hair,eyebrow";
-        private const string FULL_BODY = "fullbody";
-        private const string FULL_BODY_XR = "fullbody-xr";
-        private const string HALF_BODY = "halfbody";
         private const string PARTNER = "partner";
         private const string DATA = "data";
         private const string BODY_TYPE = "bodyType";
@@ -66,7 +63,7 @@ namespace ReadyPlayerMe.AvatarCreator
             var response = await authorizedRequest.SendRequest<Response>(
                 new RequestData
                 {
-                    Url = $"{RPM_AVATAR_V2_BASE_URL}/templates",
+                    Url = $"{RPM_AVATAR_V2_BASE_URL}/templates?{BODY_TYPE}={CoreSettingsHandler.CoreSettings.BodyType.GetDescription()}",
                     Method = HttpMethod.GET
                 },
                 ctx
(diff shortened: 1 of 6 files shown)


Runtime/AvatarCreator/Scripts/WebRequests/AvatarAPIRequests.cs
```diff
@@ -66,7 +63,7 @@ public async Task<List<AvatarTemplateData>> GetAvatarTemplates()
             var response = await authorizedRequest.SendRequest<Response>(
                 new RequestData
                 {
-                    Url = $"{RPM_AVATAR_V2_BASE_URL}/templates",
+                    Url = $"{RPM_AVATAR_V2_BASE_URL}/templates?{BODY_TYPE}={CoreSettingsHandler.CoreSettings.BodyType.GetDescription()}",
```
HarrisonHough, line 66: is it possible that we might use this body type parameter somewhere else? 
If so we could just make a function on Core settings like 
CoreSettings.GetBodyTypeParameter 
that would return this 
$"bodyType={bodyType.GetDescription}"

Runtime/AvatarCreator/Scripts/WebRequests/AvatarAPIRequests.cs
```diff
@@ -83,7 +80,7 @@ public async Task<AvatarProperties> CreateFromTemplateAvatar(string templateId,
             var payloadData = new Dictionary<string, string>
             {
                 { nameof(partner), partner },
-                { nameof(bodyType), GetBodyTypeValue(bodyType) }
+                { nameof(bodyType), bodyType.GetDescription() }
```
HarrisonHough, line 83: does this CreateFromTemplateAvatar function really need a BodyType parameter now? shouldn't it be coming from core settings? 🤔

Runtime/Core/Scripts/Extensions/EnumExtensions.cs
```diff
@@ -24,5 +25,15 @@ public static class EnumExtensions
 
             return default;
         }
+
+        public static string GetDescription(this Enum value)
```
HarrisonHough, line 29: we actually already have this function in the BodyShapeSelectionElement.cs class. However I think that a better location for it is here as you have it. So can you remove the other one from BodyShapeSelection and make it use this one? 
Screenshot 2024-06-18 at 8 30 58 am

HarrisonHough (changes requested):
looks good overall. The only thing is that we already have a GetEnumDescription function, see my comments on that. 
I also had some other questions in the comments.

Runtime/AvatarCreator/Scripts/WebRequests/AvatarAPIRequests.cs
rk132: We are still storing the avatar bodyType within the core in some places. So I don't think that's a good way to go about it for now.

HarrisonHough: okay makes sense lets leave as it is then.

HarrisonHough (approved):
Looks good, nice work!

PR #306, opened 2024-07-05
Feature/shader override update

HarrisonHough:
## Description

-   This PR reworks the shader override functionality to support multiple property types, (Previously it was only textures). This enables the proper transfer of correct values like metallic and roughness

## How to Test

-  Add a shader override to an Avatar Config
- Enter in correct target property names (these can be found by selecting your shader and looking at the property names)
- assign the config in RPM Settings window
- Load an avatar from RPM Avatar loader window
- Check that values properly mapped

## Checklist

-   [ ] Changelog is updated.

### DIFF
--- a/Editor/Core/Scripts/UI/CustomEditors/AvatarConfigEditor.cs
+++ b/Editor/Core/Scripts/UI/CustomEditors/AvatarConfigEditor.cs
@@ -38,13 +38,14 @@ namespace ReadyPlayerMe.Core.Editor
         private VisualElement root;
         private Action textureChannelChanged;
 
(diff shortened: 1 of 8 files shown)


rk132 (commented):
Values are not updated in the shader

rk132 (approved):
Looks good

rk132:
> Looks good

One improvement regarding warnings is that when the field is empty, don't show warnings

HarrisonHough:
> > Looks good
> 
> One improvement regarding warnings is that when the field is empty, don't show warnings

yes good point will add that.

PR #310, opened 2024-07-12
fix: add bones, if bones are not set

rk132:
## Description

## How to Test

## Checklist

-   [ ] Changelog is updated.


HarrisonHough (approved):
Thanks for fixing this. I think I remember that previously I worked around this by assigning a mesh in the prefab then unassigned the mesh and left the root bone and it seemed to keep all the bones. But this approach should be better, more null safe at least

PR #313, opened 2024-08-26
feat: preserve property Id from asset data

ericserrat:
[SRV-1333](https://ready-player-me.atlassian.net/browse/SRV-1333)

## Description

-   Preserve the Id from the IAsset interface, so when stripped in mobile it does not remove it

## How to Test

-   Build a sdk version for android using stripping level high, try to run it in the android emulator

[SRV-1333]: https://ready-player-me.atlassian.net/browse/SRV-1333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Runtime/AvatarCreator/Scripts/Interfaces/IAssetData.cs
+++ b/Runtime/AvatarCreator/Scripts/Interfaces/IAssetData.cs
@@ -1,7 +1,10 @@
+using UnityEngine.Scripting;
+
 namespace ReadyPlayerMe.AvatarCreator
 {
     public interface IAssetData
     {
+        [Preserve]
         public string Id { get; set; }
         public AssetType AssetType { get; set; }
     }


HarrisonHough (changes requested):
Nice work on this! 
However, I would actually suggest putting the [Preserve] attribute on the whole class rather than just on the `Id` property to make sure that everything gets preserved. 
EG
```cs
using UnityEngine.Scripting;

{
    [Preserve]
    public interface IAssetData
    {
        public string Id { get; set; }
        public AssetType AssetType { get; set; }
    }
}
```

HarrisonHough:
We could alternatively to make use of a linker file instead of using the preserve attribute. When using a link file you can preserve an entire namespace. EG something like this

```xml
<linker>
  <assembly fullname="Assembly-CSharp">
    <namespace fullname="ReadyPlayerMe.AvatarCreator" preserve="all"/>
  </assembly>
</linker>
```
This would preserve every class that is a part of ReadyPlayerMe.AvatarCreator, this may be overkill but it might prevent weird issues where certain stripping settings will strip random parts of our SDK code. 

Overall as long as the preserve tag works it doesn't really matter which approach. We should in any case make sure QA does thorough testing of this with different stripping settings on different platforms to ensure it works.

ericserrat:
> Nice work on this! However, I would actually suggest putting the [Preserve] attribute on the whole class rather than just on the `Id` property to make sure that everything gets preserved. EG
> 
> ```cs
> using UnityEngine.Scripting;
> 
> namespace ReadyPlayerMe.AvatarCreator
> {
>     [Preserve]
>     public interface IAssetData
>     {
>         public string Id { get; set; }
>         public AssetType AssetType { get; set; }
>     }
> }
> ```

Yep, I will prefer it like this, but tried it, and did not work.

ericserrat:
> We could alternatively to make use of a linker file instead of using the preserve attribute. When using a link file you can preserve an entire namespace. EG something like this
> 
> ```
> <linker>
>   <assembly fullname="Assembly-CSharp">
>     <namespace fullname="ReadyPlayerMe.AvatarCreator" preserve="all"/>
>   </assembly>
> </linker>
> ```
> 
> This would preserve every class that is a part of ReadyPlayerMe.AvatarCreator, this may be overkill but it might prevent weird issues where certain stripping settings will strip random parts of our SDK code.
> 
> Overall as long as the preserve tag works it doesn't really matter which approach. We should in any case make sure QA does thorough testing of this with different stripping settings on different platforms to ensure it works.

As discussed in huddle, seems that this is also not working for us.
For now we're going to leave it with the [Preserve] in the property

HarrisonHough (approved):
Thanks for doing that, weird that the other approach didn't work though. 

All good though.

PR #314, opened 2024-08-27
Remove Response and specify the type needed to avoid extra allocation coming from DownloadHandler.Text

abaza121:
## Description

There is an issue with avatar loader when cache is disabled this is meant to fix it by avoiding the call to downloadHandler.Text when attempting to push it.

## How to Test

-  Attempt to load avatar with caching disable and observe a spike in profiler
-  With this PR observe no spike is happening.

## Checklist

-   [ ] Changelog is updated.


HarrisonHough (approved):
Looks good to me, nice catch!

Issue #323, opened 2024-11-04
VoiceHandler memory access out of bounds on WebGL


johanqlic:
**Describe the bug**
Copy of #320, after updating to version 7.3.1 it seems the bug is back.

**Files**

I'm using an empty project with no modifications, only an imported character and a tiny mp3 file.

**To Reproduce**

1. Import a RPM character
2. Add the VoiceHandler component
3. Set AudioProvider to "AudioClip" on that component
4. Assign a random audio clip (I had the same error on different audio files)
5. Build to WebGL and play the audio (via canvas button for example, or play on awake)

**Expected behavior**

There should be no memory exception. When playing the audio manually without the VoiceHandler on a differen unity node produces no error.

**Screenshots**

![image](https://github.com/user-attachments/assets/2009753f-3bda-4b62-ad0b-5227c53abc6e)

**Desktop (please complete the following information):**
 - Ready Player Me Core version: 7.3.1
 - Ready Player Me WebView version: ?
 - glTFast version: 6.0.1 
 - Unity Editor version: 2022.3.40f1
 - Render Pipeline and version: 14.0.11
 - Operating System: MacOS
 - Platform: any browser
 
additionally (if significant for the bug):

 - Device: Macbook Pro M1 Pro
 - OS: MacOS 15.1
 - For WebGL: any Browser

PR #324, opened 2024-11-18
Fix VoiceHandler memory access out of bounds on WebGL

mennetorelli:
## Description

This should fix issue #323, where for some AudioClip lengths the GetAmplitude method tried to read data from the clip even if the remaining samples were 0.

## How to Test

- add VoiceHandler component on an avatar
- set build target to webGL and run the built application
- let the AudioSource referenced by the VoiceHandler reproduce AudioClips with various lengths

## Checklist

- [ ] Tests written or updated for the changes.
- [ ] Documentation is updated.
- [ ] Changelog is updated.


baaskoen:
@mennetorelli Thanks for yor effort! I tested this locally, and I noticed that there is still a small percentage to error, but I cant reliably reproduce it (it errors now like 1 in 100 audio files).

But this is already a huge improvement.

